Issue #38, opened 2020-06-18
Reformulate release pipelines


diocas:
We should build *and* publish our extensions automatically using github actions.

In the extension example #29, building the python binaries already builds the JS extensions automatically. So we can publish both to pypi (maybe to conda forge) and to npm.
Example on how to push to pypi: https://github.com/swan-cern/mkdocs-swan/blob/master/.github/workflows/build_publish.yml

What I expect this process to do:

1. Bump Version
    - First we need to have a way to automatically bump versions both for js and python (maybe we can create the python version file during the build process by reading package.json).
    - There are some libraries for this that also generate a changelog and create a tag in git. We can call this command locally and push to master.
1. Build extension and create release
    - When receiving a tag (created manually in the previous step), we `cd` inside the corresponding extension (since we have a monorepo, our tags need to contain the extension name) and build the python wheel. This will also build the js extensions, as mentioned above.
1. Push extension to pypi and npm
1. Notify the pipeline in the docker image to automatically bump the installed version to the new one.

For bumping the version, we could also do it automatically when we merge to master (we keep intermediate developments in a separate branch). There are actions to do this for package.json that even take commands from the PRs: https://github.com/marketplace/actions/bump-versions, https://github.com/marketplace/actions/bump-everywhere, etc

Issue #34, opened 2020-06-18
Open a Gallery notebook directly from inside Lab


diocas:
@etejedor I was planing to replace even the nbextension with swan-cern/gallery#1.
So, when you click in the "picture" icon in the top bar, you a new tab opens but in that website instead of the page we have today. What do you think?

diocas:
This has been somewhat done.. We open the gallery from Lab, as part of SwanHelp, but we still need to have an option to open the file from the iframe directly in Lab.``

diocas:
Assigning to Omar for now, maybe it will change.

Issue #37, opened 2020-06-18
Add SWAN branding to Lab interface


diocas:
And make it consistent in JupyterHub as well

rccern:
https://github.com/omazapa/SwanTheme/pull/1

PR #41, opened 2020-06-18
SwanHelp in Lab

### LINKED ISSUE #29: Create example of lab extension
To demonstrate how other extensions should be organized.

diocas:
This is SwanHelp as a Lab extension.
Closes #29 #30 

## Model for extension development
I've created this to be used as a model for all other extensions in the repo. The base organization comes from the [Jupyterlab cookiecutter](https://github.com/jupyterlab/extension-cookiecutter-ts)
I've demoted the nbextension to a subfolder and made the lab extension the main code. But both live in the same place (as they should...).
All extensions can also be distributed using the corresponding Python module. So, after doing `pip install swanhelp`, the lab extension will also be installed and we will be able enable the nbextension. This simplifies the installation process.
While not automatic yet (#38), this code is also ready to be pushed to pypi and npm.

## The help extension in lab
This extension will create a subgroup under "help" with link to the documentation (opened inside an iframe), community and SNow.
I expect to add the Gallery link also in this subgroup, but this will come from the SwanGallery extensio (#34)

### Images
image
image
image


diocas:
@prasanthkothuri @etejedor Let me know if you approve this organisation model.

etejedor:
Very nice @diocas ! My comments:

- I like the subgroup under "Help", I actually had the same thought in a conversation with Omar today, before seeing your proposal, so :+1: 
- The iframe with the help looks nice, but I guess this means we are not preserving the upper blue bar and the bottom copyright area in the whole interface (otherwise there would be a duplication).

Perhaps it is a good time to start discussing (even if grosso modo) how we would like SWAN to look like in JupyterLab. This would also determine how Omar is going to develop the extension to enable environments in projects (LCG releases, experiment stacks).

diocas:
@etejedor I have a ticket for that: #37 .
I don't believe we were ever going to put the footer. But the header, I would still like to have it. I have also thought for a while about replacing the header with a sidebar.. But at this moment I don't know if it will be easy/possible.

diocas:
@etejedor can we merge this?

PR #42, opened 2020-06-18
Hdfsbrowser to Lab

### LINKED ISSUE #31: Migrate HdfsBrowser to Lab

diocas:
This code is the one that was in out gitlab jupyterlab repo and was adapted according to the model from #29.
The change in the server extension requires a change in the way we configure it. Our current code has some values hardcoded and this new one allows for the configuration to be set in the jupyter config file.

Closes #31


omazapa:
Hi @diocas 
I am doing some tests also with the extensions 
I did this small example with a dialog to select SCRAM and CMSSW versions https://github.com/omazapa/jupyterlab-cmssw

but if we want to have the server in the same extension I am doing the tests with this one 
https://github.com/omazapa/jupyter_swan

The original example is https://github.com/jupyterlab/extension-examples/tree/master/advanced/server-extension

it would be good to have our template to work with an extension that is working with sever/frontend at the same time.


Cheers
Omar.

diocas:
@omazapa this extension has a working server, lab AND nbextension in one single place and working together. The example extension SwanHelp doesn't have a server extension but still has Python code, so it's the same.

diocas:
@prasanthkothuri can we merge this and look into the configs that we need to add in the image?

Issue #43, opened 2020-06-29
Allow new paths in the download (open with swan)


diocas:
Allow EOS paths that live in EOS projects and the new Gallery website (https://swan-gallery.web.cern.ch/)

PR #44, opened 2020-06-30
Urls support

omazapa:
code for the issue #43

### DIFF
--- a/SwanContents/swancontents/filemanager/proj_url_checker.py
+++ b/SwanContents/swancontents/filemanager/proj_url_checker.py
@@ -83,6 +83,7 @@ def check_url(url):
                      url.startswith('https://raw.githubusercontent.com') or \
                      url.startswith('https://root.cern.ch') or \
                      url.startswith('https://root.cern') or \
+                     url.startswith('https://swan-gallery.web.cern.ch') or \
                      url.startswith(CERNBoxPrefix) or \
                      url.startswith(CERNBoxPrefixTesting) or \
                      url.startswith(EOSUserPrefix) or \


SwanContents/swancontents/filemanager/proj_url_checker.py
```diff
@@ -14,7 +14,7 @@
 CERNBoxPrefixTesting = 'https://cernboxwebpreview.cern.ch/index.php/s'
 EOSUserPrefix = 'file://eos/'
 LocalPrefix = 'local:'
-EOSUserRE = '/eos/(docker/|up2u/)?(user/[a-z]|home-[a-z])/([a-z0-9]+)'
+EOSUserRE = '/eos/(docker/|up2u/)?(user/[a-z]|home-[a-z]|project-[a-z])/([a-z0-9]+)?([a-z0-9-a-z0-9]+)'
```
diocas, line 17: The paths with home-X and project-X are not the main ones (and might even get removed in the future), so you also need to support `project/[a-z]`.


I don't understand why you changed `([a-z0-9]+)` to `([a-z0-9]+)?([a-z0-9-a-z0-9]+)`. What's the reason?
And would this work with a path that contains spaces? And if the url is encoded (space becomes %20)?

diocas: I'm not even sure if we should just allow `/eos`, which would work with `/eos/cms/etc` for example. @etejedor @ebocchi do we have any security issue? The permissions are already managed by EOS, so it should be fine.

omazapa: > I don't understand why you changed `([a-z0-9]+)` to `([a-z0-9]+)?([a-z0-9-a-z0-9]+)`. What's the reason?
> And would this work with a path that contains spaces? And if the url is encoded (space becomes %20)?

because some projects  has dashes in the names
ex:
re.search('/eos/(docker/|up2u/)?(user/[a-z]|home-[a-z]|project-[a-z])/([a-z0-9]+)?([a-z0-9-a-z0-9]+)', 'https://cern.ch/swanserver/cgi-bin/go?projurl=file://eos/project-a/acc-models/public/ps/2021/scenario
   ...: s/ad/0_injection/MADX_example_ps_inj_ad.ipynb')[0]

output '/eos/project-a/acc-models'

without it 

re.search('/eos/(docker/|up2u/)?(user/[a-z]|home-[a-z]|project-[a-z])/([a-z0-9]+)', 'https://cern.ch/swanserver/cgi-bin/go?projurl=file://eos/project-a/acc-models/public/ps/2021/scenarios/ad/0_injection/MA
   ...: DX_example_ps_inj_ad.ipynb')[0]

output: /eos/project-a/acc'   <- this is wrogn

diocas: true and we should accept all. Let's just check that the file is in `/eos` and that's it.

omazapa: something like this 
EOSUserRE = '/eos/'
??

diocas: In that case we don't need to use it as a regular expression, just checking if the path is `file:/eos/`

diocas: Ahhh, by looking at the latest commits you did I realized why I added this in the first place... To make it easier to get the username/project name... So we might still need the regex for that.. Just add the `/project-X/` or `/project/X/`. And to check if the download if valid or not, just check if it starts on EOS.

But did you test your changes? I'm pretty sure what you did will crash..

grupocolav: Not, I did not test it, I was just looking the results importing python re  and passing 
an url to see if the regex was working.
Let me know what should I do.
I can to revert the last commit if you want.

diocas: Sorry, I just checked the code (which I didn't remember) and I don't understand why it would fail to download a file from any EOS instance already...  Including projects. 
The user was saying that it worked for him, right?
When you told me it was failing, I said it was because of this regex but it doesn't seem to be true.. In the code we only check if the the url starts with `file:/eos` to allow it. And then we check if the eos path points to the user home folder or somewhere else.

diocas: So, I recommend you double check if sharing something from a project (for example, the swan eos project) works. If so, we just add the new url for the gallery website and close this.

omazapa: I will check it

PR #70, opened 2020-09-03
Pip installable extensions

diocas:
This PR looks big, but I basically changed the folders and files around so that they all have the same structure of the already merged SwanHelp extension. This means that all of them are ready to be pushed to PyPi and, if applicable, they have a Jupyterlab extension together with server and nb extensions. The lab extension is empty for the moment, but this way we have everything ready to start coding without worrying about how to build and package.

Other changes:
* Added compatibility with Notebook 6.1.*
* Removed SwanGallery extension: we now rely only on swan-gallery.web.cern.ch
* Disabled the publish to NPM, since the majority of our Lab extensions are still empty (but everything is packaged in the python module, so they are in reality being installed already)

I will make a PR for system user to install them with this new way.

Should I tag all versions as v1.0.0? (we were not doing versioning right before, so there's nothing to compare too.. And they have been in prod for a while, so I don't see a problem in using v1)


etejedor (approved):

.github/workflows/swan-ci-ca.yml
```diff
@@ -86,7 +91,7 @@ jobs:
         #Full path with package name, version and organization
         NPM_PKG_PATH=build/lib/$NPM_PKG_NAME/labextension/swan-cern-$NPM_PKG_NAME-$NPM_PKG_VERSION.tgz
 
-        npm publish $NPM_PKG_PATH --access public
+        npm publish $NPM_PKG_PATH --access public --dry-run
```
omazapa, line 94: @diocas 
with --dry-run the package will not be uploaded

omazapa: on sparkmonitor you have README.md and README copy.md

diocas: > Temporary: disable pushing to NPM for now since our JupyterLab extensions are empty

diocas: ups, I'll remove the copy..

PR #80, opened 2020-09-26
sparkmonitor: Fix for Spark 3.0.1: use attemptNumber instead of attemptId

pwais:
Attempting to use `sparkmonitor` (either the original one or the fork in this repo) with Spark 2.4.7 works (at least in this tree https://github.com/pwais/oarphpy/tree/v0.0.4 ), but with Spark 3.0.1 one runs into this error:
```
INFO:SparkMonitorKernel:Client Connected ('127.0.0.1', 43708)
2020-09-26 06:54:06,315 ERROR util.Utils: uncaught error in thread spark-listener-group-shared, stopping SparkContext
java.lang.IllegalAccessError: class sparkmonitor.listener.JupyterSparkMonitorListener tried to access private method 'int org.apache.spark.scheduler.StageInfo.attemptId()' (sparkmonitor.listener.JupyterSparkMonitorListener and org.apache.spark.scheduler.StageInfo are in unnamed module of loader 'app')
        at sparkmonitor.listener.JupyterSparkMonitorListener.stageInfoToJSON(CustomListener.scala:172)
        at sparkmonitor.listener.JupyterSparkMonitorListener$$anonfun$onJobStart$4.apply(CustomListener.scala:204)
        at sparkmonitor.listener.JupyterSparkMonitorListener$$anonfun$onJobStart$4.apply(CustomListener.scala:203)
        ...
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:62)
        at org.apache.spark.scheduler.AsyncEventQueue.org$apache$spark$scheduler$AsyncEventQueue$$dispatch(AsyncEventQueue.scala:100)
        at org.apache.spark.scheduler.AsyncEventQueue$$anon$2.$anonfun$run$1(AsyncEventQueue.scala:96)
        at org.apache.spark.util.Utils$.tryOrStopSparkContext(Utils.scala:1319)
        at org.apache.spark.scheduler.AsyncEventQueue$$anon$2.run(AsyncEventQueue.scala:96)
```

While it appears `attemptId` was made private years ago ( https://github.com/apache/spark/commit/32365f8177f913533d348f7079605a282f1014ef#diff-e3e46893527ae54411257d6fb1d73964 ), nevertheless you can see in the javadocs that `attemptId` is gone in Spark 3.x and they want people to use `attemptNumber`:
 * Spark 2.4.7: https://spark.apache.org/docs/2.4.7/api/java/org/apache/spark/scheduler/StageInfo.html
 * Spark 3.x (latest): https://spark.apache.org/docs/latest/api/java/org/apache/spark/scheduler/StageInfo.html

This change includes the following fixes:
 * Use `attemptNumber` instead of `attemptId`.  I'm not sure how (if at all) the semantics differ, but I imagine the end-user impacts will be minimal.  In my own experience debugging spark jobs, if I have to be very nitty gritty I'll try to get the raw logs from the executor or history server.
 * Sets up the scala listener to build against Spark 3.0.1 (which appears to require updating to scala 2.12.  For whatever reason, I could not get the listener to build against Spark 2.1.1 as-is.  Building against Spark 2.7.4 does appear to work.

So far I've only tested this manually, not sure what sort of automated testing might be available.

### DIFF
--- a/SparkMonitor/scalalistener/CustomListener.scala
+++ b/SparkMonitor/scalalistener/CustomListener.scala
@@ -169,7 +169,7 @@ class JupyterSparkMonitorListener(conf: SparkConf) extends SparkListener {
     val submissionTime: Long = stageInfo.submissionTime.getOrElse(-1)
 
     (stageInfo.stageId.toString ->
-      ("attemptId" -> stageInfo.attemptId) ~
+      ("attemptId" -> stageInfo.attemptNumber) ~
       ("name" -> stageInfo.name) ~
       ("numTasks" -> stageInfo.numTasks) ~
       ("completionTime" -> completionTime) ~
@@ -217,7 +217,7 @@ class JupyterSparkMonitorListener(conf: SparkConf) extends SparkListener {
     // so that we can display stage descriptions for pending stages:
     for (stageInfo <- jobStart.stageInfos) {
       stageIdToInfo.getOrElseUpdate(stageInfo.stageId, stageInfo)
-      stageIdToData.getOrElseUpdate((stageInfo.stageId, stageInfo.attemptId), new StageUIData)
+      stageIdToData.getOrElseUpdate((stageInfo.stageId, stageInfo.attemptNumber), new StageUIData)
     }
     val name = jobStart.properties.getProperty("callSite.short", "null")
     val json = ("msgtype" -> "sparkJobStart") ~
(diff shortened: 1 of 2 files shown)


pwais:
Here is how I tested:
 * See this `Dockerfile` content for how I check out and build this PR: https://github.com/pwais/oarphpy/blob/ebe136a5b81c50aa5dfd789d4c77cd91099ecb55/docker/full.Dockerfile#L134 
 * You can run a test notebook then using `docker run -it --rm --net=host oarphpy/full:0.1.0 jupyter notebook --allow-root --ip="*" --port=9999`  (and connect to the jupyter instance it starts)

Here are screen shots from doing that on my machine showing that sparkmonitor is working as expected:
Screen Shot 2020-09-26 at 01 35 12
Screen Shot 2020-09-26 at 01 34 58

diocas:
Thanks for your contribution, @pwais !
I'll ask a colleague (@prasanthkothuri) to review your PR and test everything, and I'll create a new release with the fix.

prasanthkothuri:
@pwais Thanks for your contribution, I tested and it works fine with spark 3, however we want to have sparkmonitor working with spark 3 (scala 2.12) and spark 2 (scala 2.11), so we are thinking of making it like https://issues.apache.org/jira/browse/SPARK-32017

then user will install as below

SCALA_VERSION=2.12.11
pip install sparkmonitor

we will do the necessary changes and merge this

diocas:
Seems that this change will come from #84 
So I'm closing it.
Thanks for the contribution!

PR #84, opened 2020-10-15
Spark3fix

prasanthkothuri:


SparkMonitor/sparkmonitor/kernelextension.py
```diff
@@ -186,7 +187,10 @@ def configure(conf):
     log.info(os.environ["SPARKMONITOR_KERNEL_PORT"])
     conf.set("spark.extraListeners",
              "sparkmonitor.listener.JupyterSparkMonitorListener")
-    jarpath = os.path.abspath(os.path.dirname(__file__)) + "/listener.jar"
+    if "2.11" in get_spark_version():
```
diocas, line 190: seems a bit unnecessary to execute a command 2 times (we should store the version result in a var).
Also, there's no else, so this might crash?

diocas (changes requested):
Simple changes

SparkMonitor/scalalistener/CustomListener.scala
```diff
@@ -169,7 +169,7 @@ class JupyterSparkMonitorListener(conf: SparkConf) extends SparkListener {
     val submissionTime: Long = stageInfo.submissionTime.getOrElse(-1)
 
     (stageInfo.stageId.toString ->
-      ("attemptId" -> stageInfo.attemptId) ~
+      ("attemptId" -> stageInfo.attemptNumber) ~
```
pwais, line 172: Thank you !!! 🎉

SparkMonitor/.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.0.0
```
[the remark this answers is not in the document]
prasanthkothuri: ok, reverting now

SparkMonitor/sparkmonitor/kernelextension.py
prasanthkothuri: good poing, adding variable. today all of spark is with scala 11 or 12, even otherwise it doesn't crash, extension has nothing to show

diocas (approved):

PR #91, opened 2021-01-29
Fix python3.6 compatibility

dolfinus:
Hello.

According to setup.py file of SparkMonitor, minimal supported version of Python is 3.5.
But when I tried to run it on python3.6, an exception has been raised:
```
  File "/opt/anaconda2/envs/jh-py36/env/lib64/python3.6/subprocess.py", line 403, in run
    with Popen(*popenargs, **kwargs) as process:
TypeError: __init__() got an unexpected keyword argument 'capture_output'
```
`capture_output` argument was added  to python 3.7+ and it does not exist on python 3.6. In this PR I fixed this compatibility issue.

But after fixing it I've received another exception:
```
  File "/opt/anaconda2/envs/jh-py36/env/lib64/python3.6/site-packages/sparkmonitor/kernelextension.py", line 185, in configure
    log.info("SparkConf Configured, Starting to listen on port:", str(port))
Message: 'SparkConf Configured, Starting to listen on port:'
Arguments: ('56189',)
```
It is because the string format in the log line has no substitutions, but some value is passed into it. This error is fixed too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@
 !SwanContents/swancontents/js/
 SparkMonitor/scalalistener/project/
 SparkMonitor/scalalistener/target/
-SparkMonitor/sparkmonitor/listener.jar
+SparkMonitor/sparkmonitor/listener*.jar
 SparkMonitor/sparkmonitor/nbextension/*
 .idea/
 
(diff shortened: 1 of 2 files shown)


diocas:
Thank you for your commit! (And sorry for taking so long to reply).
@prasanthkothuri @krishnan-r can you have a look?

krishnan-r (approved):

PR #92, opened 2021-02-02
implemented is_project to support jupyterlab

omazapa:
Hi guys,

This is required for our the JupyterLab extensions,
this allows to know when I am inside of project, but I changed the way I am sending the information,
type in the json is changed from project to directory and a new tag is created "is_project".

This is because the integration SwanContents into the Jupyterlab FileBrowser should be done in this way.
https://discourse.jupyter.org/t/custom-project-extension/6598/3

Omar.

### DIFF
--- a/SwanContents/swancontents/filemanager/swanfilemanager.py
+++ b/SwanContents/swancontents/filemanager/swanfilemanager.py
@@ -84,6 +84,9 @@ class SwanFileManager(SwanFileManagerMixin, LargeFileManager):
 
         if parent_project and not parent_project == 'invalid':
             model['project'] = parent_project
+            model['is_project'] = True
+        else:
+            model['is_project'] = False
 
         return model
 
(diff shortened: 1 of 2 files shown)


diocas:
@omazapa can you also take the oportunity to fix this in the js code?

I see 2 options:
* You change `notebooklist.js` to do `model.is_project` instead of `model.type == 'project'`, but then you need to figure out how to display the Projects icon in the list (should not be that hard). (ps we might also need to change things in the sharing extension, not sure..)
* You intercept the calls to the API in `contents.js` and transform the results from `is_project == True` to `type == project`.

omazapa:
Hi @diocas 

I did the fixes suggested by you and I manage to fix even the icons.
take a look at this line [(here)](https://github.com/swan-cern/jupyter-extensions/pull/92/commits/bd9819996ad7643dd8680c0da3a40fe12de2df12#diff-aefb0752e1ff347256929fdb13f4879179050ecb678a38ab00b594aa2a0b5073R251)
I did the test on sciencebox because in swan-qa002 we dont have enough space to build the docker images to test the extensions.

couple of snapshots with the icons
image
image

diocas (approved):

PR #104, opened 2021-02-19
Fixes for Contents, NotebookViewer and OauthRenew

### LINKED ISSUE #20: Rendering of readme.md files inside project sub-directories
Requested by Michał:

I am writing to propose a new feature. Currently, SWAN renders readme.md file only for the main project folder, while for the subdirectories it is not done. I would like to add more description on what notebook (from a subdirectory) to choose for a given analysis. Hence, having a dedicated readme.md file would be useful. In other words I want to get the gitlab behavior:
1. Main project directory: https://gitlab.cern.ch/LHCData/lhc-sm-hwc 
2. Project sub-directory: https://gitlab.cern.ch/LHCData/lhc-sm-hwc/-/tree/master/600A

### LINKED ISSUE #100: SwanOauthRenew: configurable tokens/outputs
We should add a configuration where we provide functions to resolve from a token (and we can have many, thanks to swan-cern/jupyterhub-extensions#29) to an output (a file).

For example, for EOS we need to have a file with a specific format (`oauth2:eos_token:issuer_url`), but we might want to have a separate file for CERNBox which just has the token.

I suggest we pass a dict to the config, with the file path as key and a method, that receives the tokens list and returns an output, as the value.

### LINKED ISSUE #101: Downloading zip files from gallery fails
If we have notebooks that require data, the open in swan points to a zip file, but we do not accept that extension.

### LINKED ISSUE #102: Error opening a notebook shared with me
This functionality relies on nbconvert, which got updated to the latest version and that contains breaking changes with the old configurations. We need to downgrade or fix the configuration. (check our mkdocs script, which had the same problem in the past)

diocas:
Sorry for grouping but this way it's quicker to me.

This PR:
* closes #20 
* closes #100 
* fixes #101 
* fixes #102

It also does small fixes to be easier to deploy locally and (temporarily) fixes a problem created by #92 

The fixes to SwanShare will probably come with another PR.

### DIFF
--- a/SwanOauthRenew/README.md
+++ b/SwanOauthRenew/README.md
@@ -1,6 +1,8 @@
 # SwanOauthRenew
 
-Extension that refreshes the Oauth token 
+This is a server extension that fetches the oAuth tokens available in the "user" endpoint of the JupyterHub API (1) and stores them in files that can be accessed by other processes (like the EOS client). It can be configured to take any token from the auth state dictionary and write it with any content format.
+
+(1) For now, this functionality is exclusive to SwanHub (our JH wrapper). It does not work on vanilla JH, as the auth state is only available to admins.
 
 ## Requirements
 
@@ -20,4 +22,12 @@ Configure the server extension to load when the notebook server starts
  jupyter serverextension enable --py --user swanoauthrenew
 ```
 
-This extension updates the oauth2 token in the file pointed by the environment variable OAUTH2_FILE.
+Then is necessary to configure (in the jupyter config file) the files that need to be written, from where the tokens are coming from and how the file content should look like. Like so:
+
+```python
+c.SwanOauthRenew.files = [
+        ('/tmp/swan_oauth.token', 'access_token', '{token}'),
+        ('/tmp/cernbox_oauth.token', 'exchanged_tokens/cernbox-service', '{token}'),
+        ('/tmp/eos_oauth.token', 'exchanged_tokens/eos-service', 'oauth2:{token}:auth.cern.ch')
+    ]
+```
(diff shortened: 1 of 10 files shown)


SwanOauthRenew/README.md
```diff
@@ -1,6 +1,8 @@
 # SwanOauthRenew
 
-Extension that refreshes the Oauth token 
+This is a server extension that fetches the oAuth tokens available in the "user" endpoint of the JupyterHub API (1) and stores them in files that can be access by other processes (like the EOS client). It can be configured to take any token from the auth state dictionary and write it with any content format.
```
etejedor, line 3: `that can be accessed`

etejedor: Can you elaborate a bit on the purpose of this extension after the modifications? It does not refresh the oauth token anymore? What's the use of storing them in files? Are the other processes that read the files use the tokens for authentication themselves?

etejedor (dismissed):
Added some comments on the OAuth renew part, the other commits are simple and look ok. And +1 for showing the readmes in subfolders :)

SwanOauthRenew/README.md
diocas: Before this extension was looking for a single token (which was in a known location inside a dict), and storing it in a file (configured via env variables). We only had one token back then, which was wrong, and now we have more tokens, 1 per service we need to connect to.

What the new extension does is removing all the hardcoded stuff. So now you can say where to look in the dict, and where to store the final file. Basically, in the first version it was hardcoded for EOS, now it's configurable for any service we need. 

But the _raison d'être_ of the extension is still the same: pick a token that JH renewed, and store it in a file where running processes can look.

PR #110, opened 2021-03-16
ITHADOOP-1020: Configure shuffle for Spark 3.x on K8s

krishnan-r:
This PR moves the options in the Spark3Shuffle bundle, to be hardcoded on the connector for Spark 3 running on K8s (cloud containers cluster). (These options need to be set always for Spark3 on K8s for dynamic resource allocation to work without an external shuffle service)

Ideally this config could live near `gitlab/db/hadoop-confext/conf/etc/analytix/spark3.analytix/spark-defaults.conf`, but these options are specific to Spark on K8s only.


LucaCanali (approved):
LGTM

PR #111, opened 2021-03-16
[ITHADOOP-1024] Remove check on spark.jars.packages for Spark on K8S

LucaCanali:
Spark on K8S version 2.4 does not support --packages (confing parameter spark.jars.packages).
The Spark connector in SWAN throws an error, signaling this.
Spark 3.x supports the use of spark.jars.packages (in client mode).
As we have now moved to Spark 3.0.1 in LCG 99 and Spark 3.1.1 in future LCG 100, this updates the SWAN connector removing the check so that we can use the new functionality.


rccern (approved):
LGTM @diocas can you merge this?

diocas:
@rccern gave you permissions. feel free to tag a new version

PR #176, opened 2021-08-30
SwanContents LAB3 updates

omazapa:
Hi Guys,

some updates for SwanContents, 
to be consistent with all the components at the moment the project can only exist in the
SWAN_projects folder, others folder outside this folder with the file .swanproject is not a project.

Also updated notebook version, compatible with lab 3.0.x

Tested in swan-spare001  please take a look there.

### DIFF
--- a/SwanContents/setup.py
+++ b/SwanContents/setup.py
@@ -55,7 +55,6 @@ setup_args = dict(
         "License :: OSI Approved :: GNU Affero General Public License v3",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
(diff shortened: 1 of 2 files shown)


diocas:
The topic introduced by this PR still needs to be discussed. Is it really that big of a deal for the frontend extension to force the usage of Projects inside this folder?
My point at the meeting was that if the backend tells the frontend that you are a Project/are inside a Project, you show the new Launcher options. Meaning that this folder should not be hardcoded in the frontend code, it should instead rely on this information that comes with the contents api response.

Thoughts, @omazapa @etejedor ?

(ps: keep this PR with the projects functionality and remove anything else that is not related)

etejedor:
I believe the contents of this PR are outdated w.r.t. our last discussion in the SWAN meeting? We don't have anymore a projects view, we can't restrict anymore projects to be in a certain folder.

diocas:
I don't think we finished the discussion because Omar said he depended on it. That why I said above that this still needs to be discussed.
I would like to be sure this dependency cannot be removed.

PR #180, opened 2021-08-31
Add Header Element SWAN JupyterLab

krishnan-r:
A small proof of concept that adds a top bar to the interface. We can perhaps iterate on the design.

- Removes the default logo and puts it on the top bar.

image


diocas:
Yes, the current Help implementation adds buttons to the help page, the community, to SNow and we even have the gallery.
If we enable binder we will have to enable the multiple (named) server functionality from JH, which allows users to start more than one session. So yes, multiple containers per user.
The multiple UIs will take some time, but if it's just a config json we can already implement it as it should be easy. I had other ideas to put them in some other places, but I didn't like the mockup (not did I have better ideas). If you have suggestion I can try to create some mockups.
If you see my mockups there are no SWAN logo in the bottom. All of that, and the blue toolbar + grey-er UI, was part of a SWAN template that we decided was not worth having (especially given the maintenance). So the idea is to keep everything upstream and just add the header.

krishnan-r:
Thanks for the inputs, Diogo, Enric,

- I like the idea of having a menu for multiple named servers, and moving logout to the menu.

- Can we still keep a big red shutdown button outside the menu in the top left, I was thinking that if this were prominently visible, it would encourage users to shutdown their session after use, saving us resources (both in our cluster and external like active connections to hadoop/spark). I've observed that the current "Change Configuration" button is something many users discover when we explicitly tell them.

- I have to check about using gravatar, and how that works with cern emails and privacy concerns etc. Also it may not work in network isolated machines. (Or if there is a gitlab api like for mattermost)

krishnan-r:
image
@diocas I've implemented the basic layout as per your designs. I'll add the notifications and multiple servers as a separate PR. Can you review this again and merge

SwanHeader/setup.py
```diff
@@ -0,0 +1,88 @@
...
+        "License :: OSI Approved :: BSD License",
+        "Programming Language :: Python",
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Framework :: Jupyter",
+        # These classifiers will be valid when https://github.com/pypa/warehouse/pull/9882 gets merged
+        # "Framework :: Jupyter :: JupyterLab",
```
diocas, line 62: ? why is this commented?

SwanHeader/src/header.tsx
```diff
@@ -0,0 +1,118 @@
...
+            transformOrigin={{
+              vertical: 'top',
+              horizontal: 'right'
+            }}
+          >
+            <MenuItem button onClick={onClickLogout} dense divider>
+              <ListItemText primary="Logout" />
+            </MenuItem>
+            <MenuItem button onClick={onClickShutdown} dense>
+              <ListItemText primary="Shutdown" />
```
diocas, line 111: I know this was raised by you recently, but the production nomenclature is still `change configuration`. We should change this once we enable multiple servers + the new launcher, but for now I would keep it consistent.

SwanHeader/src/header.tsx
```diff
@@ -0,0 +1,118 @@
...
+import ListItemText from '@material-ui/core/ListItemText';
+import Avatar from '@material-ui/core/Avatar';
+
+export const Header = (props: {
+  hubPrefix?: string;
+  hubUser?: string;
+  baseUrl: string;
+}): React.ReactElement => {
+  const userName = props.hubUser || 'swanuser';
+  const shutdownUrl = (props.hubPrefix || props.baseUrl) + 'home?changeconfig';
```
diocas, line 19: Shouldn't this always be `hubPrefix`?  Isn't `baseUrl` the base for the user container?

SwanHeader/style/base.css
```diff
@@ -0,0 +1,86 @@
+#swan-header {
+  --swan-color-dark-1: #09263c;
```
diocas, line 2: Fine for now. Eventually we should create a separate extension (used as dependency) with these tokens and the logos.

SwanHeader/style/base.css
```diff
@@ -0,0 +1,86 @@
...
+  height: 100%;
+  padding: 3px 12px;
+  transition: height 2s linear;
+  display: flex;
+  align-items: center;
+}
+
+.sw-header .sw-nav-item:hover, .sw-header .sw-nav-item-selected {
+  border-top: 3px solid transparent;
+  border-bottom: 3px solid transparent;
```
diocas, line 58: Please make the borders 2px instead

SwanHeader/style/base.css
```diff
@@ -0,0 +1,86 @@
...
+  --swan-border-radius: 1px;
+
+  height: 40px !important;
+  min-height: 40px;
+}
+
+.sw-header .sw-swan-logo {
+  width: 40px;
+  height: 40px;
+  background-image: url(swan_cloud.png);
```
diocas, line 17: Wondering if we should take the logo gives by `/hub/logo` instead...

diocas (commented):
Some comments.
Please add the bump2version to make the release procedure consistent.
Can you remove the top border of the dropdown menu?

SwanHeader/.prettierignore
```diff
@@ -0,0 +1,5 @@
+node_modules
```
omazapa, line 1: the same question here, should we have a .prettier* files for each extension?

SwanHeader/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
```
omazapa, line 1: same as above

SwanHeader/setup.py
```diff
@@ -0,0 +1,88 @@
...
+    long_description_content_type="text/markdown",
+    packages=setuptools.find_packages(),
+    install_requires=[],
+    zip_safe=False,
+    include_package_data=True,
+    python_requires=">=3.6",
+    platforms="Linux, Mac OS X, Windows",
+    keywords=["Jupyter", "JupyterLab", "JupyterLab3"],
+    classifiers=[
+        "License :: OSI Approved :: BSD License",
```
omazapa, line 53: AGLP

SwanHeader/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "clean:all": "jlpm run clean:lib && jlpm run clean:labextension",
+    "eslint": "eslint . --ext .ts,.tsx --fix",
+    "eslint:check": "eslint . --ext .ts,.tsx",
+    "install:extension": "jlpm run build",
+    "watch": "run-p watch:src watch:labextension",
+    "watch:src": "tsc -w",
+    "watch:labextension": "jupyter labextension watch ."
+  },
+  "dependencies": {
+    "@jupyterlab/application": "^3.1.0",
```
omazapa, line 48: is this compatible with jupyterlab 3.0.x?

SwanHeader/pyproject.toml
```diff
@@ -0,0 +1,17 @@
+[build-system]
+requires = ["jupyter_packaging~=0.10,<2", "jupyterlab~=3.1"]
```
omazapa, line 2: same here jupyter 3.1

SwanHeader/setup.py
krishnan-r: This was in the older version of the cookiecutter template, back when I generated the code from it. Will remove.

SwanHeader/package.json
```diff
@@ -0,0 +1,81 @@
...
+  ],
+  "styleModule": "style/index.js",
+  "jupyterlab": {
+    "extension": true,
+    "outputDir": "swanheader/labextension"
+  },
+  "jupyter-releaser": {
+    "hooks": {
+      "before-build-npm": [
+        "python -m pip install jupyterlab~=3.1",
```
omazapa, line 76: same here jupyter 3.1

SwanHeader/src/header.tsx
krishnan-r: True, let's keep it consistent, I will rename it

omazapa (commented):
Hi @krishnan-r,  thanks for the extension, some of my comments.

the .bumpversion.cfg  is missing as well.

SwanHeader/package.json
krishnan-r: it worked in our latest image with 3.0.x, but I will still downgrade it to prevent any issues

SwanHeader/.eslintignore
```diff
@@ -0,0 +1,5 @@
+node_modules
```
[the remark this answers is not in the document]
omazapa: Agree! I think is better.

SwanHeader/style/base.css
krishnan-r: I'm not sure how to do this, webpack tries to parse any relative URLs (/hub/logo) I add here and errors with file not found.

diocas: Not sure either.. Leave it. But the best would be to take it from there, **if exists**, or from your local asset.

diocas (approved):

PR #186, opened 2021-10-11
Disable console progress and INFO logs

rccern:
sparkConnector 1.1.1 has this fix already


diocas:
And please commit the version change to master as well.

diocas:
Clarification: there was the need to update an older version, so we branched. tag was ok.

PR #188, opened 2021-11-12
add config validate_cert 

akhilputhiry:
Some times we don't want to validate the SSL certificates of name nodes.
Added an extra configuration to ignore ssl validation

### DIFF
--- a/HdfsBrowser/hdfsbrowser/serverextension.py
+++ b/HdfsBrowser/hdfsbrowser/serverextension.py
@@ -3,7 +3,7 @@ import re
 import os
 import traceback
 
-from traitlets import Unicode, Long, Float
+from traitlets import Unicode, Long, Float, Bool
 from traitlets.config import Configurable
 
 from notebook.utils import url_path_join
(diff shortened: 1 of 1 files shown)


rccern:
Hi,
Thanks for the PR!
We are already developing a mechanism to handle SSL enabled clusters, but we are implementing the check at runtime (reading hdfs-site.xml) instead of with configuration as it's easier for us in this period when we are progressively enabling SSL in our clusters.
For the second part (validate_cert), can you rebase once our commit for using SSL clusters is in? (I will ping you here)

Can you send the changes without isort and black? We will enforce style through CI at some point, but we are not there yet.

akhilputhiry:
Sure @rccern Thanks

rccern:
@akhilputhiry  do you still want to merge this?

akhilputhiry:
@rccern if possible, yes

rccern (approved):

rccern:
merged, thank you for the PR @akhilputhiry !

PR #229, opened 2022-04-04
SwanDask: wrapper to dask-labextension

diocas:
This simple wrapper allows us to run Dask not as a server extension, but as a separate process where we can configure the environment. I don't think we can get simpler, and less maintenance free, than this.

For now, the environment is taken from the python3 kernel (this could be configurable at some point). 
**I haven't tested it with the LCG stack, so we might need to fix some details** (add some local path in the env, if we serve the dask extension locally, or maybe change the way we call the process, since the current approach will probably use the local python binary instead of the lcg). But these are very minor issues.

This extension requires a to-be-released version of `jupyter-server-proxy`. The latest doesn't allow empty bodies on PUT requests, which the Dask extension does when creating a cluster. This is already fixed in master, so it's a matter of waiting for a new release or compile it.

@etejedor can you test and check the details I've mentioned?

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,13 @@
+[flake8]
+ignore = E501, W503, E402
+builtins = c, get_config
+exclude =
+    .cache,
+    .github,
+    docs,
+    setup.py
+enable-extensions = G
+extend-ignore =
+    G001, G002, G004, G200, G201, G202,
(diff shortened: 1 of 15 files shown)


rccern:
can you run precommit here and force push the results?

etejedor (approved):

PR #237, opened 2022-04-12
Separate PortAllocator from SparkConnector

etejedor:
Even though `SparkConnector` is the only SWAN component that needs to allocate ports at the moment, the future support for Dask HTCondor and its need for port allocation motivates the separation of `PortAllocator` in its own extension, which will centrally manage ports for any SWAN component that needs them.

### DIFF
--- a/SparkConnector/setup.py
+++ b/SparkConnector/setup.py
@@ -36,10 +36,6 @@ package_data_spec = {
 
 data_files_spec = [
     ("share/jupyter/labextensions/@swan-cern/sparkconnector", lab_path, "**"),
-    ("etc/jupyter/jupyter_server_config.d",
-     "jupyter-config/jupyter_server_config.d", "sparkconnector.json"),
-     ("etc/jupyter/jupyter_notebook_config.d",
-     "jupyter-config/jupyter_notebook_config.d", "sparkconnector.json"),
 ]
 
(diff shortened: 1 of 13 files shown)


diocas (approved):

SparkConnector/setup.py
```diff
@@ -69,6 +69,7 @@
     install_requires=[
         "jupyterlab~=3.0",
         "bs4",
+        "swanportallocator",
```
krishnan-r, line 72: You might need to move `jupyter-config/` from `data_files_spec` into the new extension

etejedor: Missed that, thanks!

SparkConnector/setup.py
```diff
@@ -36,10 +36,6 @@
 
 data_files_spec = [
     ("share/jupyter/labextensions/@swan-cern/sparkconnector", lab_path, "**"),
-    ("etc/jupyter/jupyter_server_config.d",
```
etejedor, line 39: Is this fine @krishnan-r ? (plus the addition to the `setup.py` of `SwanPortAllocator`)

krishnan-r: yes

krishnan-r (approved):
lgtm

PR #252, opened 2022-05-17
SwanProjects extension

krishnan-r:
- Uses https://jupyter-client.readthedocs.io/en/latest/provisioning.html to dynamically load environments defined in a folder when a kernel starts
- Forks `@jupyterlab/launcher` to display a dialog to edit environment options for a folder
- Server extension with handlers to create and edit metadata files 

Needs https://github.com/swan-cern/systemuser-image/pull/81

Based on https://github.com/swan-cern/jupyter-extensions/pull/209


SwanProjects/swanprojects/config.py
```diff
@@ -0,0 +1,33 @@
+from traitlets import Unicode
+from traitlets.config import Configurable
+import pathlib
+import os
+
+class SwanProjectsConfig(Configurable):
```
diocas, line 6: Please add documentation to this. There are some variables that are not clear what they are.

SwanProjects/swanprojects/utils.py
```diff
@@ -0,0 +1,233 @@
...
+                **pass_through_env,
+            },
+        )
+        stdout, stderr = await process.communicate()
+        if process.returncode != 0:
+            self.log.error(
+                f"RETURNCODE: {process.returncode}, STDOUT:\n {stdout}\n, STDERR: \nf{stderr}\n"
+            )
+            raise Exception(
+                f"Error generating environment with output {stdout}, {stderr}, {process.returncode}"
```
diocas, line 172: This is sent to the users, right? Maybe not worth to put all this detail.

SwanProjects/swanprojects/utils.py
```diff
@@ -0,0 +1,233 @@
...
+            self.contents_manager.delete_file(user_script_file_path)
+        else:
+            model = {
+                "name": self.swan_config.project_file_name,
+                "path": user_script_file_path,
+                "type": "file",
+                "content": user_script,
+                "format": "text",
+            }
+            self.contents_manager.save(model, user_script_file_path)
```
diocas, line 221: Why do you use this method? There're other, lower level ones, that you can use and don't require you to set all of this model stuff.
(same applies to other places you use it)

SwanProjects/swanprojects/swanprovisioner.py
```diff
@@ -0,0 +1,51 @@
...
+            self.log.error(message)
+            raise Exception(message)
+
+        kwargs["env"] = dict()
+        with open(environment_file_path) as f:
+            for line in f.readlines():
+                name, value = line.rstrip("\n").split("=", 1)
+                kwargs["env"][name] = value
+
+        self.kernel_spec.argv[0] = shutil.which(
```
diocas, line 39: Please document

SwanProjects/swanprojects/swanprovisioner.py
```diff
@@ -0,0 +1,51 @@
...
+            os.path.basename(self.kernel_spec.argv[0]),
+            path=kwargs["env"].get("PATH", None),
+        )
+        if self.kernel_spec.argv[0] is None:
+            raise Exception("This kernel is not avaialble in this enviornment.")
+        return await super().pre_launch(**kwargs)
+
+    def _get_project_root(self, path: str) -> Optional[Path]:
```
diocas, line 47: Why the duplication?

diocas (commented):
Didn't review the ts code yet (either way, you know that way better than anyone else). But in the meantime, I'm adding some comments/nitpicks. 
In general, I would add a bit more comments, as it would make things more understandable (even if we can understand after spending some time).

At the end, I just remembered that some of this logic still lives in the contents manager, no? Meaning that there's code shared among both, so maybe we should re-use the utils there, for example.
Or we no longer need it?

SwanProjects/README.md
```diff
@@ -0,0 +1,23 @@
...
+- A kernel provisioner that dynamically customizes the enviornment of a kernel from metadata files in a folder. 
+
+## Requirements
+
+JupyterLab~=3.0
+
+## Install & Configure
+
+- Install the package with pip
+- Configure `SWAN_DEFAULT_ENV_FILE` for the default environment for folders without an environment
```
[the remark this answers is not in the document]
diocas: (sorry, even myself had to go back to try to understand what I mean with this comment)
What I mean is, instead of setting this as an extra variable, we just push the file we generate on startup on the "default" folder.

SwanProjects/swanprojects/config.py
```diff
@@ -0,0 +1,33 @@
...
+
+    environment_pass_through_variables = ["LD_LIBRARY_PATH"]
+
+    default_environment_file_path = os.environ.get("SWAN_DEFAULT_ENV_FILE", "")
+
+    userscript_file_name = Unicode(
+        ".userscript", config=False, help="User script file name."
+    )  # type: str
+
+    forbidden_project_folders = ["", "SWAN_projects"]
```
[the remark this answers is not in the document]
diocas: regex in general, since we would prob not put a wildcard to avoid catching the full path.

SwanProjects/swanprojects/utils.py
```diff
@@ -0,0 +1,233 @@
...
+        ----------
+        path : str
+            A path to the project.
+
+        Returns
+        -------
+        path: Dict || None
+            Project information in a dictionary or a empty dictionary in case of error.
+        """
+        if self.has_project_file(folder_path):
```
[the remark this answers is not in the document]
diocas: (reason: the network calls might be expensive-_ish_)

PR #262, opened 2022-08-04
Swan gallery

DYasser:


SwanGallery/src/handler.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  let response: Response;
+  try {
+    response = await ServerConnection.makeRequest(requestUrl, init, settings);
+    }   
+    catch (error) {
+    throw new ServerConnection.NetworkError(<TypeError>error);
+  }
+
+  let data: any = await response.text();
```
krishnan-r, line 37: Can this be simplified with `response.json()` directly?

Also the `catch` in line 43 swallows the error returning the text, is that intentional?

SwanGallery/src/__tests__/SwanGallery.spec.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Example of [Jest](https://jestjs.io/docs/getting-started) unit tests
+ */
+
+describe('SwanGallery', () => {
```
krishnan-r, line 5: Can we remove this file if it is not used? (If you have time, you can add some tests here otherwise.)

SwanGallery/setup.py
```diff
@@ -0,0 +1,111 @@
...
+     "jupyter-config/server-config", "SwanGallery.json"),
+    # For backward compatibility with notebook server
+    ("etc/jupyter/jupyter_notebook_config.d",
+     "jupyter-config/nb-config", "SwanGallery.json"),
+]
+
+long_description = (HERE / "README.md").read_text(encoding="utf8")
+
+version = (
+    pkg_json["version"]
```
krishnan-r, line 41: Can you configure bump2version similar to [other extensions in this repo](https://github.com/swan-cern/jupyter-extensions/blob/master/HdfsBrowser/.bumpversion.cfg).

The python package is built and pushed to pypi with a new version automatically using a [release pipeline](https://github.com/swan-cern/jupyter-extensions/blob/c309c0caf7494befbf9ed8ef9b6beed1e09a3f00/.github/workflows/release.yml#L34) documented in the [README](https://github.com/swan-cern/jupyter-extensions#create-a-release). So we don't have to manually upgrade versions.

https://pypi.org/project/bump2version/

README.md
```diff
@@ -6,6 +6,7 @@ Repository that stores all the Jupyter extensions for SWAN.
 * [SparkConnector](SparkConnector) - Helper to connect to CERN's Spark Clusters
 * [SwanContents](SwanContents) - Contents Manager for Jupyter with Projects functionality and SWAN templates
 * [SwanDask](SwanDask) - Wrapper to run dask_jupyterlab as an external process
+* [SwanGallery](SwanGallery) - Service to help users with SWAN notebooks by providing a set of notebooks and project examples easy to access and download.
```
krishnan-r, line 9:
```suggestion
* [SwanGallery](SwanGallery) - Extension to help users get started with SWAN by providing a gallery of example notebooks and projects that can be easily viewed and downloaded from within JupyterLab
```

SwanGallery/.github/workflows/build.yml
```diff
@@ -0,0 +1,149 @@
+name: Build
```
krishnan-r, line 1: [TODO/LATER] These pipelines do not run as they are in a subfolder. They also need to be adapted to the existing pipelines already in the `.github` top level folder.

SwanGallery/package.json
```diff
@@ -0,0 +1,119 @@
...
+    "clean:lib": "rimraf lib tsconfig.tsbuildinfo",
+    "clean:lintcache": "rimraf .eslintcache .stylelintcache",
+    "clean:labextension": "rimraf SwanGallery/labextension",
+    "clean:all": "jlpm clean:lib && jlpm clean:labextension && jlpm clean:lintcache",
+    "eslint": "jlpm eslint:check --fix",
+    "eslint:check": "eslint . --cache --ext .ts,.tsx",
+    "install:extension": "jlpm build",
+    "lint": "jlpm stylelint && jlpm prettier && jlpm eslint",
+    "lint:check": "jlpm stylelint:check && jlpm prettier:check && jlpm eslint:check",
+    "prettier": "jlpm prettier:base --write --list-different",
```
krishnan-r, line 46: Can you run these lint tools (`npm run prettier` etc) and update the PR.

We are not currently running these from CI.

SwanGallery/SwanGallery/handlers.py
```diff
@@ -0,0 +1,200 @@
+import json
```
krishnan-r, line 1: If I understand correctly, these python files are simply copied from `SwanContents` right?

Once we refactor/figure out how to use `SwanContents` locally, we can get rid of these files right?


If not, i.e if there is some extra logic added on top, can you add a comment on the top of the file and describe what it changes

SwanGallery/SwanGallery/tests/test_handlers.py
```diff
@@ -0,0 +1,13 @@
+import json
+
+
+async def test_get_example(jp_fetch):
+    # When
+    response = await jp_fetch("SwanGallery", "get_example")
```
krishnan-r, line 6: Can we remove this file, (or even better add some tests if we have time)

krishnan-r (changes requested):
Hi Yasser,

Thanks for putting this PR together. Here is an initial review. I have added some comments, mostly suggestions. 

Some of the comments may be for future/later. But I have just added them for when we pick this work up later. So feel free to skip larger changes.

SwanGallery/src/index.ts
```diff
@@ -0,0 +1,78 @@
...
+  console.log('JupyterLab extension SwanGallery is activated!');
+  var flag = true;
+  const command: string = 'swangallery:open';
+  app.commands.addCommand(command, {
+    label: `SWAN Gallery`,
+    icon: swanGalleryIcon,
+    execute: () => 
+    { 
+      let content = new IFrame({
+        sandbox: ['allow-scripts', 'allow-forms', 'allow-same-origin', 'allow-modals', 'allow-downloads']
```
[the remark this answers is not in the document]
DYasser: No, I will remove it.

SwanGallery/src/index.ts
```diff
@@ -0,0 +1,78 @@
...
+        window.addEventListener('message', event => {
+          requestAPI<any>(event.data, 'notebook')
+          .then(data => {
+            app.commands.execute('filebrowser:open-path', {
+              path: data.path,
+              showBrowser: false,
+            });
+          })
+          .catch(reason => {
+            console.error(
```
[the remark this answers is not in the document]
krishnan-r: Perhaps "Failed to download notebook"

SwanGallery/src/handler.ts
DYasser: You are right, I simplified it and works fine. 

``` typescript
  let data: any;

  try {
    data = await response.json();
  } catch (error) {
    console.log('Not a JSON response body.', error);
  }
```
As for the error, it was giving me more details that way when I was testing. I will make it use `error` now.

SwanGallery/src/__tests__/SwanGallery.spec.ts
DYasser: The test files were automatically generated, I will remove it.

diocas:
Closing has it was superseded with another PR that took your changes and fixed what was pending. Thanks for contributing!

PR #271, opened 2022-11-09
[SPARK-8] fix broken links and update descriptions in the window "Spark clusters connection

LucaCanali:
Fix broken links to spark context and spark session description + propose minor change to the description of the WebUI and Spark Metrics Dashboard links


krishnan-r (approved):
Thanks, LGTM

PR #284, opened 2022-11-14
Fix bumpversion conflict in SparkConnector package.json

krishnan-r:


etejedor:
This PR implements a more strict pattern matching when bumping the version of SparkConnector, in a similar way to what was done for the SWAN charts. It also reverts two wrong modifications that were made as a result of the old pattern matching.

etejedor (approved):

PR #287, opened 2022-11-16
Disable console progress for Spark on K8S

LucaCanali:
Disable console progress as it would be printed in the notebook as of ipython 6 for Spark on K8S.
This configuration is already there for Spark on YARN/HADOOP.


krishnan-r (approved):
LGTM, thanks

PR #288, opened 2022-11-18
Add extension to show a popup when a notebook save fails

krishnan-r:
image

- This ideally could be part of the SwanContents extension, but the nbextensions in SwanContents are not installed, but loaded from the page templates directly, so I added this as a separate extension, so that it can be installed and enabled separately.


diocas (changes requested):
Hey, please add this directly in the template (`notebook.html`). I don't think is worth to build another extension for this.

krishnan-r:
Closing in favor of #289

PR #289, opened 2022-11-19
SwanContents: add token authorization to requests

diocas:
Handle failed notebook saves

Replaces #288

The idea of this PR is to put the notification directly in the page instead of an extra extension.
I also added the logic to do requests using the token (like lab does), even if for now we don't have the token available (it should eventually...).

Tested on my dev jupyter-server branch, but it should be working... Can you please give it a try?


SwanContents/swancontents/__init__.py
```diff
@@ -21,6 +21,11 @@ def _jupyter_nbextension_paths():
                 src="js",
                 dest="swancontents",
                 require="swancontents/notebooklist"),
+            dict(
+                section="tree",
```
krishnan-r, line 25:
```suggestion
                section="notebook",
```

krishnan-r (approved):

PR #304, opened 2023-01-19
SwanGallery JupyterLab Extension

krishnan-r:
Continuation of work done in https://github.com/swan-cern/jupyter-extensions/pull/262

Requires 
- https://github.com/swan-cern/mkdocs-swan/pull/2
- https://github.com/swan-cern/swan-charts/pull/128
- https://github.com/swan-cern/systemuser-image/pull/95

PR #307, opened 2023-02-26
Merge SwanGallery with SwanHelp extension

krishnan-r:
This PR takes the work done in https://github.com/swan-cern/jupyter-extensions/tree/update_lab_3 (to update the SwanHelp extension to JupyterLab v3) and merges the work done in #304 (SwanGallery JupyterLab extension) into a single extension for Help Menus and Examples Gallery.

Replaces #304 as discussed in the comments there.

Requires 
- https://github.com/swan-cern/mkdocs-swan/pull/2
- https://github.com/swan-cern/swan-charts/pull/128
- https://github.com/swan-cern/systemuser-image/pull/95


krishnan-r:
Future TODO / Improvements

- [ ] Make the mkdocs site not collapse the sidebar inside the iframe when the width of the iframe is small. This makes the page confusing to navigate
- [X] Add a spinner for downloading notebooks

SwanHelp/src/handler.ts
```diff
@@ -0,0 +1,43 @@
+import { URLExt } from '@jupyterlab/coreutils';
+import { ServerConnection } from '@jupyterlab/services';
+
+/**
+ * Calls /api/contents/fetch/ with the url to download,
```
etejedor, line 5: We should probably mention that this helper function is part of the Gallery functionality. Also the name of the file (`handler.ts`) is too generic and it should change to something more gallery-related so that by looking at the name one knows what's inside? Perhaps `gallery_download_handler.ps` ?

SwanHelp/src/index.ts
```diff
@@ -1,25 +1,44 @@
...
+
+      app.commands.execute('filebrowser:open-path', {
+        path: response.path,
+        showBrowser: false
+      });
+      dialog.reject();
+    } catch (error) {
+      dialog.reject();
+      showDialog({ title: "Failed to download notebook", buttons: [Dialog.okButton()] });
+      console.error(
```
diocas, line 29: This error message is a bit strange. We might have the extension running and fail for smth else. But this is not necessarily user visible, but I would still put some generic message.

SwanHelp/src/index.ts
```diff
@@ -1,25 +1,44 @@
...
+      );
+    }
+  });
+}
 
 const defaultConfigs = {
   help: "https://swan.docs.cern.ch/",
   community: "https://cern.ch/swan-community",
   support: "https://cern.service-now.com/service-portal/function.do?name=swan",
-  gallery: "https://cern.ch/swan-gallery"
```
diocas, line 11: Why don't we keep just 1 config and use this default here? If the extension is now called swan help, what's the reason behind spliting into 2 config files?

SwanHelp/src/index.ts
```diff
@@ -105,17 +131,38 @@ async function activate(
...
+  if (launcher) {
+    launcher.add({
+      command: CommandIDs.gallery,
+      category: 'Other',
+      rank: 1
+    });
+  }
 }
+
+const extension: JupyterFrontEndPlugin<void> = {
```
diocas, line 161: Why we moved this to the end?

SwanHelp/swanhelp/__init__.py
```diff
@@ -1,15 +1,31 @@
...
+def _jupyter_labextension_paths():
+    return [{
+        'src': 'labextension',
+        'dest': data['name']
+    }]
+
+
+# Server extension
+def _jupyter_server_extension_points():
+    return []
```
diocas, line 22: Let's keep the comment.

diocas (commented):
Just some details and some questions. Otherwise, I think we can merge.

SwanHelp/src/index.ts
diocas: Or does this make it easier somehow to configure from the k8s land?

etejedor (approved):

PR #318, opened 2023-08-03
Use the WebUI for Spark on YARN instead of using Spark history server.

LucaCanali:
The use of the Spark history server was a workaround introduced for SWAN on k8s, this appears not to be necessary any more, the WebUI should be preferred as it has more functionality.

### DIFF
--- a/SparkConnector/nbextension/src/extension.js
+++ b/SparkConnector/nbextension/src/extension.js
@@ -1010,7 +1010,7 @@ SparkConnector.prototype.get_html_connected = function (config, error) {
     if (config && config.sparkhistoryserver) {
         var historyserverURL = $('<a>').attr('href', config.sparkhistoryserver).attr('target','_blank').text('here')
         html.find('.success-history-text')
-            .text('Spark WebUI: History Server is available ')
+            .text('Spark WebUI is available ')
             .append(historyserverURL)
     }
 
--- a/SparkConnector/nbextension/src/extension.js
+++ b/SparkConnector/nbextension/src/extension.js
@@ -1007,11 +1007,11 @@ SparkConnector.prototype.get_html_connected = function (config, error) {
             .append(metricsURL)
     }
 
-    if (config && config.sparkhistoryserver) {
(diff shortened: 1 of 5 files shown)


SparkConnector/sparkconnector/configuration.py
```diff
@@ -399,15 +399,12 @@ def get_spark_session_config(self):
                                               '&var-UserName=' + self.get_spark_user() + \
                                               '&var-ApplicationId=' + app_id
 
-            # determine the history server URL depending on the selected resource manager (yarn, k8s, local etc)
-            historyserver_url = self._get_sc_config(
-                'spark.yarn.historyServer.address',
+            # determine the WebUI URL for Spark on YARN
```
etejedor, line 402: Can we add here in the comment that what we obtain is the URLs of the Hadoop NNs?

SparkConnector/sparkconnector/configuration.py
```diff
@@ -399,15 +399,12 @@ def get_spark_session_config(self):
...
+                'spark.org.apache.hadoop.yarn.server.webproxy.amfilter.AmIpFilter.param.PROXY_URI_BASES',
                 wait=True
             )
-            historyserver_protocol = 'https' if self._get_sc_config('spark.ssl.historyServer.enabled') else 'http'
-
-            ui_url = f'{historyserver_protocol}://{historyserver_url}/history/{app_id}'
-            if ui_url:
-                conn_config['sparkhistoryserver'] = ui_url
+            if webui_urls:
+                conn_config['sparkhistoryserver'] = webui_urls.split(',', 1)[0]
```
etejedor, line 408: This configuration option should not be called sparkhistoryserver anymore, right? We need to rename this option here.

SparkConnector/nbextension/src/extension.js
```diff
@@ -1010,7 +1010,7 @@ SparkConnector.prototype.get_html_connected = function (config, error) {
     if (config && config.sparkhistoryserver) {
         var historyserverURL = $('<a>').attr('href', config.sparkhistoryserver).attr('target','_blank').text('here')
```
etejedor, line 1011: And we also need to rename here.

etejedor (approved):

PR #329, opened 2023-10-12
SwanHeader: automatic conversion to Lab 4 and code fixes

diocas:


etejedor:
@diocas so you used some tool to automatically migrate the extension? From the things you did manually, is there something worth being in its own commit for better future tracking of changes?

diocas:
I didn't accept the output of the automated process blindly (meaning that I made 1 or 2 changes to the default behavior). But don't think it deserves a separate commit.

PR #330, opened 2023-10-12
SwanHelp: compatibility with Lab 4

diocas:
Kept the old build configuration to build both lab and nbextension as independent code


SwanHelp/swanhelp/_version.py
```diff
@@ -1,19 +1,4 @@
...
-            pass
-
-    raise FileNotFoundError('Could not find package.json under dir {}'.format(HERE))
-
-__version__ = _fetchVersion()
-
+# This file is auto-generated by Hatchling. As such, do not:
+#   - modify
+#   - track in version control e.g. be sure to add to .gitignore
+__version__ = VERSION = '2.0.2'
```
etejedor, line 4: Why 2.0.2 here?

PR #334, opened 2023-10-17
Sparkconnector

diocas:


SwanPortAllocator/swanportallocator/__init__.py
```diff
@@ -4,7 +4,7 @@ def _jupyter_nbextension_paths():
     # Empty to avoid error when automatically trying to enable all nbextensions
     return []
 
-def _jupyter_server_extension_paths():
+def _jupyter_server_extension_points():
```
etejedor, line 7: Does this change match the commit message?

SparkConnector/README.md
```diff
@@ -2,83 +2,92 @@
...
 
-Note: You will need NodeJS to install the extension.
+To install the extension, execute:
 
 ```bash
 pip install sparkconnector
-jupyter nbextension install sparkconnector --py
-jupyter nbextension enable  sparkconnector --py
-jupyter lab build
+jupyter nbclassic-extension install sparkconnector --py
```
etejedor, line 25: Is @PMax5 already using this new way of installing in his Alma9 image? Just making sure.

SparkConnector/setup.py
```diff
@@ -63,7 +63,7 @@
     cmdclass= cmdclass,
     packages=setuptools.find_packages(),
     install_requires=[
-        "jupyterlab~=3.0",
+        "jupyterlab>=4.0.0,<5",
```
etejedor, line 66: Here JupyterLab is required to install and to build?

SwanPortAllocator/swanportallocator/__init__.py
diocas: Yes.

SparkConnector/README.md
diocas: Not yet. I still need to review that part as I hope many of those can be (and are already) automated. And we need to check if we need to disable some as well (i.e the spark connector for lab should not load automatically as well).

SparkConnector/setup.py
diocas: I'm not understanding the question.

diocas: Ah sorry, yes, you need lab to build things as well. Don't remember why, but that is part of their cookiecutter recipe.

etejedor (approved):

PMax5 (approved):

PR #335, opened 2023-10-18
SwanContents with Jupyter Server

diocas:
Ports the contents extension+manager to the Server backend. This means the creation of `SwanClassic`, which is a wrapper of `nbClassic` with our personalisations (we need to disable `nbclassic`, though, to avoid conflicts with the urls).
It's a bit debatable if SwanClassic should be split into a separate extension or not. I'm open to suggestions.

Did some refactoring as well to start integrate the new CERNBox (CS3) as an option and to rely as much as possible on upstream code and re-implement just the bare minimum to support SWAN Projects.

I'll start looking into the new CERNBox integration soon, but this as it is now should allow us to spawn an Alma 9 image with all dependencies as latest and have the same look and feel (and functionality) of the current prod system.

Since SwanNotebookViewer is no longer used anywhere else besides in here, I merged it into SwanContents (otherwise, it would've been harder to update as it would require other configs for theme files and such).

I've also simplified the configuration, as most of things are now enabled by default when pip install this package.


SwanContents/swancontents/js/utils.js
```diff
@@ -0,0 +1,21 @@
+define([
+    'base/js/utils'
+], function (utils) {
+
+    // Wrap upstream promising_ajax to append the authorization header
```
etejedor: Can the commit message explain a bit more why we now put this token in the authorization header? What is this token? What are the requests that now will have this token? Why wasn't this there before?

SwanContents/swancontents/__init__.py
```diff
@@ -1,36 +1,22 @@
 from ._version import __version__ 
-from .filemanager import *
-from .handlers import *
+# from .filemanager import *
```
etejedor: Can we just remove these two lines?

SwanContents/swancontents/swanclassic/handlers/share.py
```diff
@@ -1,10 +1,16 @@
...
+from tornado import web, gen
 
+from jupyter_server.base.handlers import JupyterHandler
+from jupyter_server.extension.handler import (
+    ExtensionHandlerMixin,
+    ExtensionHandlerJinjaMixin
+)
+from jupyter_server.base.handlers import path_regex
+from jupyter_server.utils import url_path_join, url_escape, ensure_async
+from swancontents.filemanager.fileio import swan_sharing_folder
```
etejedor: There's no problem with an _explicit_ relative import, as you had initially, and actually you should use it if you can. What is forbidden in Python3 is an _implicit_ relative import (see [this](https://docs.python.org/release/3.0.1/whatsnew/3.0.html#removed-syntax)).

SwanContents/swancontents/swanclassic/handlers/projects.py
```diff
@@ -59,9 +73,9 @@ def get(self, path=''):
                                             server_root=self.settings['server_root_dir'],
                                             projects_page=True,
                                             ))
-        elif cm.file_exists(swan_path):
+        elif file_exists:
             # it's not a directory, we have redirecting to do
-            model = cm.get(swan_path, content=False)
+            model = yield ensure_async(cm.get(swan_path, content=False))
```
etejedor: Why do you need to make the `get` a coroutine and yield on it?

SwanContents/swancontents/swanclassic/handlers/projects.py
```diff
@@ -32,23 +38,31 @@ def generate_page_title(self, path):
...
     def get(self, path=''):
         """ Get handler to cope with Projects as root folder """
         path = path.strip('/')
         cm = self.contents_manager
 
         swan_path = url_path_join('SWAN_projects', path)
 
-        if cm.dir_exists(path=swan_path):
-            if cm.is_hidden(swan_path):
+        file_exists = False
```
etejedor, line 56: Is this change of logic related to migrating to jupyter_server? What's the motivation for it?

SwanContents/swancontents/swanclassic/notebookapp.py
```diff
@@ -0,0 +1,52 @@
...
+
+from jupyter_server.transutils import _i18n
+from jupyter_server.serverapp import load_handlers
+
+from traitlets import Unicode
+import datetime
+
+
+class NotebookApp(ClassicNotebookApp):
+
```
etejedor, line 19: A bit of class docstring would not hurt (purpose of this class, etc.).

SwanContents/swancontents/swanclassic/notebookapp.py
```diff
@@ -0,0 +1,52 @@
...
+    Serves the old Jupyter UI with the SWAN configurations (i.e templates) on top""")
+
+    default_url = Unicode("/projects").tag(config=True)
+
+    static_paths = [DEFAULT_STATIC_FILES_PATH, NBCLASSIC_DEFAULT_STATIC_FILES_PATH]
+    template_paths = DEFAULT_TEMPLATE_PATH_LIST + NBCLASSIC_DEFAULT_TEMPLATE_PATH_LIST
+
+    def initialize_handlers(self):
+        handlers = []
```
etejedor, line 32: Function docstring please.

SwanContents/swancontents/swanclassic/notebookapp.py
```diff
@@ -0,0 +1,52 @@
...
+        handlers.extend(load_handlers('swancontents.swanclassic.handlers.projects'))
+        handlers.extend(load_handlers('swancontents.swanclassic.handlers.share'))
+        handlers.extend(load_handlers('swancontents.swanclassic.handlers.cernbox'))
+        self.handlers.extend(handlers)
+        super(NotebookApp, self).initialize_handlers()
+
+    def initialize_settings(self):
+        super(NotebookApp, self).initialize_settings()
```
etejedor, line 41: Docstring please.

SwanContents/swancontents/swanclassic/notebookapp.py
```diff
@@ -0,0 +1,52 @@
...
+        new_vars = self.settings['jinja_template_vars'] if 'jinja_template_vars' in self.settings else dict()
+        new_vars.update({'current_year':datetime.datetime.now().year})
+
+        if datetime.date.today().month == 12:
+            # It's Christmas time!
```
etejedor, line 46: :D How is this here? The SWANta only shows in the main page that is served by the hub, before the notebook server is even started, correct?

SwanContents/swancontents/swanclassic/templates/page.html
```diff
@@ -238,6 +264,30 @@
         });
     });
     require(['template']);
+    
+    function _remove_token_from_url() {
```
etejedor, line 269: What is this for?

SwanContents/swancontents/serverextension/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+TEMPLATE_PATH = [os.path.join(os.path.dirname(__file__), "templates")]
+
+# Automatically configure our templates
+@property
+def template_file_path(self):
+    return self.extra_template_paths + TEMPLATE_PATH + DEFAULT_TEMPLATE_PATH_LIST
+ServerApp.template_file_path = template_file_path 
+
+def _load_jupyter_server_extension(serverapp):
```
etejedor, line 19: "New handlers" is not new anymore, is it? Please some docstring that specifies the handlers that this server extension installs.

Also please at least a paragraph in the commit message that explains this move, as well as the reason to strip out a few of the handlers and where are those going.

SwanContents/swancontents/filemanager/swanfilemanager.py
```diff
@@ -337,43 +343,6 @@ def update(self, model, path):
 
         return super(LargeFileManager, self).update(model, path)
 
-    def delete(self, path, force=False):
-        """Delete a file/directory and any associated checkpoints."""
-        path = path.strip('/')
-        if not path:
-            raise web.HTTPError(400, "Can't delete root")
-        self.delete_file(path, force)
-        self.checkpoints.delete_all_checkpoints(path)
```
etejedor, line 346: So now we won't delete the checkpoints?

SwanContents/swancontents/filemanager/__init__.py
```diff
@@ -1 +1,6 @@
-from .swanfilemanager import *
\ No newline at end of file
+from .swan_eos_filemanager import SwanEosFileManager
+
+try:
+    from .swan_cs3_filemanager import SwanCs3FileManager
+except:
+    pass
```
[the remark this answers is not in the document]
etejedor: Why the try-catch in the first place?

SwanContents/swancontents/filemanager/swan_cs3_filemanager.py
```diff
@@ -0,0 +1,41 @@
+from .projects_mixin import ProjectsMixin
+from cs3api4lab import CS3APIsManager
+
+class SwanCs3FileManager(ProjectsMixin, CS3APIsManager):
```
etejedor, line 4: I am hesitating whether we should include this now, if it's just an empty shell. We could just do the refactoring to make the current filemanager the eos file manager, that is good, but the Cs3FileManager inclusion is just adding a skeleton without real functionality. This should be delayed until there is real functionality.

SwanContents/swancontents/filemanager/swan_eos_filemanager.py
```diff
@@ -0,0 +1,51 @@
...
+from .eos.fileio import SwanFileManagerMixin
+from .eos.handlers import SwanAuthenticatedFileHandler
+from .projects_mixin import ProjectsMixin
+from ..checkpoints.eoscheckpoints import EOSCheckpoints
+from traitlets import default
+import os, shutil
+
+
+class SwanEosFileManager(ProjectsMixin, SwanFileManagerMixin, LargeFileManager):
+    """ SWAN File Manager Wrapper
```
etejedor, line 12: SWAN File Manager Wrapper for content on EOS?

SwanContents/swancontents/swanclassic/handlers/notebookviewer.py
```diff
@@ -0,0 +1,79 @@
...
+except:
+    pass
+
+
+class NotebookViewerHandler(ExtensionHandlerJinjaMixin, ExtensionHandlerMixin, JupyterHandler):
+    """
+        Jupyter server extension to provide a view-only mode to open notebooks.
+        When users receive a shared project, now they can open it and use this extension
+        to load the notebooks in view-mode (without the editing options that would allow
+        the them to edit the document, even without permissions to save the changes).
```
etejedor, line 24: `the them` -> `them`

SwanContents/swancontents/filemanager/projects_mixin.py
```diff
@@ -204,16 +204,6 @@ def save(self, model, path=''):
 
         return model
 
-    def new(self, model=None, path=''):
```
etejedor: Why does `new` go away? We can just rely on the superclass' `new`? Why?

SwanContents/swancontents/filemanager/__init__.py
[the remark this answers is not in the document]
diocas: Because we don't want to install the CS3 module (which brings a couple of dependencies) if we don't strictly need to. That's why I have a try-catch: if it's installed, we can use it, if not, no harm is done to the default manager.
I don't add a log because I would have to instantiate a logger, but if you try to use it and it's not importer, I think the error should be sufficient.

SwanContents/swancontents/__init__.py
diocas: This is already removed in a later commit.

diocas: But I also cleaned this in a soon to be pushed commit

SwanContents/swancontents/swanclassic/notebookapp.py
diocas: This includes SWANta in the download page, which also shows the same loading effect as the hub.

SwanContents/swancontents/swanclassic/templates/page.html
diocas: This is code copied from upstream. Just aligning our code.

SwanContents/swancontents/filemanager/swanfilemanager.py
diocas: This was not the code we added. We added the possibility to delete a folder that contained some files inside (i.e the `.swanproject`). But with the force delete enabled, upstream now works the same way as we did (so, it still deletes checkpoints).

SwanContents/swancontents/filemanager/projects_mixin.py
diocas: Because in this commit we no longer use the type "project". Now it's a folder that contains a "project" attribute in the model. 

When Omar was here, we changed the logic only for the response. I simplified it even further to make the same logic in the creation thus removing extra logic required.

SwanContents/swancontents/swanclassic/handlers/notebookviewer.py
diocas: This commit was just a move of the file xD the text was incorrect in the code already. Should I create a different commit or combine in this?

etejedor: Just combining is fine.

SwanContents/swancontents/swanclassic/handlers/projects.py
diocas: This was part of the code changes with CS3 (I was planning to expose the full EOS namespace). I removed it for now, will come later with a different PR.

SwanContents/setup.py
```diff
@@ -36,15 +34,13 @@
     url="https://github.com/swan-cern/jupyter-extensions",
     author="SWAN Admins",
     description="SWAN Contents Manager for Jupyter",
-    long_description= long_description,
+    long_description=long_description,
```
etejedor, line 37: These are all format changes, but ok... let's just get done with it.

SwanContents/swancontents/serverextension/handlers/__init__.py
```diff
@@ -1,39 +0,0 @@
...
-import datetime
-
-def load_jupyter_server_extension(nb_server_app):
-    """ Used as a server extension in order to install the new handlers """
-
-    new_handlers = [(r"/cernbox", TreeHandler),
-                    (r"/cernbox%s" % path_regex, TreeHandler),
-                    (r"/projects", ProjectsHandler),
-                    (r"/projects%s" % path_regex, ProjectsHandler),
-                    (r"/share", ShareHandler),
```
etejedor: I also asked to explain in the commit message where handlers like this one go, since they are no longer here.

SwanContents/swancontents/serverextension/templates/page.html
```diff
@@ -0,0 +1,92 @@
...
+  </div>
+
+  {% block after_site %}
+  {% endblock after_site %}
+
+  {% block script %}
+  {% endblock script %}
+
+  <script type='text/javascript'>
+    function _remove_token_from_url() {
```
etejedor, line 66: I don't know if we discussed about this function? What does it do? Some doc would be nice?

SwanContents/setup.py
diocas: Some are, others aren't. I applied the formatter to files that were being changed.

SwanContents/swancontents/serverextension/handlers/__init__.py
diocas: They are created in a different place on the previous commit. I'm going to delete this file on the previous commit then.

SwanContents/swancontents/serverextension/templates/page.html
diocas: We discussed. It comes from upstream, it's part of aligning our files with upstream changes

etejedor: True!

diocas: Updated the commit message to state that the update means pulling stuff from upstream.

SwanContents/swancontents/swanclassic/handlers/projects.py
diocas: https://www.tornadoweb.org/en/stable/guide/coroutines.html#native-vs-decorated-coroutines

I kept the old style in nbclassic handlers (they also use it), but actually, it would be okay to use await everywhere, as we only support Python 3.9... In some places I had even messed up... I'll update everywhere (only using async await).

etejedor: Yes let's use native coroutines everywhere!

etejedor (approved):

PR #342, opened 2023-11-16
Make portallocator generic to also handle requests from the Dask lab extension

etejedor:
This changes the port allocator to:
- Handle a variable with generic computing ports, which can be requested by either the Spark connector or the Dask lab extension.
- If a process already asked for ports, do not return the same ports if the same process asks again -- this is important for the Dask lab extension to work, because on every creation of a SwanHTCondorCluster, a new port will be requested by the same process
- Be able to release ports (when a SwanHTCondorCluster is shut down, its corresponding port is released).

### DIFF
--- a/SwanPortAllocator/swanportallocator/portallocator.py
+++ b/SwanPortAllocator/swanportallocator/portallocator.py
@@ -28,6 +28,7 @@ class Errors(Enum):
 # Actions pickable to be sent between master and clients
 class Actions(Enum):
     GET_PORT = "get_port"
+    RELEASE_PORT = "release_port"
     SET_STATUS = "set_status"
 
 
@@ -81,6 +82,24 @@ class PortAllocator(threading.Thread):
 
         self.log.info('Requested ports for process %s: %s' % (process, assigned_ports))
         return assigned_ports
+
+    def release_ports(self, process, ports):
+        """
+            Release a list of ports from a process.
+        """
+
+        if process not in self.clients:
+            self.log.warn(f'Process {process} not in registered clients, not releasing ports {ports}')
+            return
+
+        stored_ports = self.clients[process]['ports']
+        for p in ports:
+            try:
+                stored_ports.remove(p)
+                self.ports_available.append(p)
+                self.log.info(f'Port {p} from process {process} has been released')
+            except ValueError:
+                self.log.warn(f'Port {p} not assigned to process {process}, not releasing port {p}')
 
     def delete_client(self, process):
         """ Delete a client from the list of processes and put its ports back in the list so that they're reused """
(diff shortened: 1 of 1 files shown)


SwanPortAllocator/swanportallocator/portallocator.py
```diff
@@ -62,35 +63,46 @@ def __init__(self, log):
     def get_ports(self, process, n):
         """
             Return 'n' available ports and assign them to 'process' pid.
-            If the 'process' already asked for ports, and the information was not recycled, return the same ports.
             Raises NoPortsException if all ports are in use.
         """
 
         if process in self.clients:
```
diocas, line 69: Single line if?

SwanPortAllocator/swanportallocator/portallocator.py
```diff
@@ -62,35 +63,46 @@ def __init__(self, log):
...
-                self.log.info('Returning same ports for process %s: %s' % (process, ports))
-                return ports
-            else:
-                n = n - len(ports)
+            stored_ports = self.clients[process]['ports']
         else:
-            ports = []
+            stored_ports = []
 
         if len(self.ports_available) < n:
```
diocas, line 74: Since we no longer change `n`, I would put this check as the first thing we do.

SwanPortAllocator/swanportallocator/portallocator.py
```diff
@@ -216,6 +228,10 @@ def send_error(content):
                             ports = self.get_ports(message['process'], message['n'])
                             send_ok(ports)
 
+                        elif message['action'] == Actions.RELEASE_PORT.value:
+                            self.release_ports(message['process'], message['ports'])
+                            send_ok(message['ports'])
```
diocas, line 233: Why do you want to know the ports that have been removed (and therefore reply back with them)? If you do, maybe `release_ports` should return them (since they might not be the same as the ones sent).

diocas: I guess it's for consistency, but I don't remember why bellow I sent the status back...

diocas (changes requested):

SwanPortAllocator/swanportallocator/portallocator.py
etejedor: You mean `stored_ports = self.clients[process]['ports'] if process in self.clients else []`

SwanPortAllocator/swanportallocator/portallocator.py
etejedor: Good point, will do.

SwanPortAllocator/swanportallocator/portallocator.py
etejedor: Yes it was for consistency. As you say the status below is also sent and discarded.

SwanPortAllocator/swanportallocator/portallocator.py
etejedor: Done.

SwanPortAllocator/swanportallocator/portallocator.py
etejedor: Should I add a commit that removes that parameter from `send_ok` both for set status and release ports?

diocas: Yes, please

etejedor: For the actually released ports, we already issue a warning if we can't release a port because it does not belong to that process, I think that is enough and anyway, as you say, we won't do anything with those ports.

diocas (approved):

PR #361, opened 2024-01-25
Configure spark.kubernetes.container.image depending on the Linux distribution

LucaCanali:
This is to add configuration for setting spark.kubernetes.container.image
The configuration is via an environemnt variable that can be set by the spawner.
This will allow to support executor containers built for different OSes (notably CC7 and Alma9)

### DIFF
--- a/SparkConnector/sparkconnector/configuration.py
+++ b/SparkConnector/sparkconnector/configuration.py
@@ -269,11 +269,6 @@ class SparkK8sConfiguration(SparkConfiguration):
         conf.set('spark.kubernetes.namespace', os.environ.get('SPARK_USER'))
         conf.set('spark.master', self._retrieve_k8s_master(os.environ.get('KUBECONFIG')))
 
-        # Configure shuffle if running on K8s with Spark 3.x.x
-        if self.get_spark_version().split('.')[0]=='3':
-            conf.set('spark.shuffle.service.enabled', 'false')
-            conf.set('spark.dynamicAllocation.shuffleTracking.enabled', 'true')
-
         # Ensure that Spark ENVs on executors are the same as on the driver
         conf.set('spark.executorEnv.PYTHONPATH', os.environ.get('PYTHONPATH'))
         conf.set('spark.executorEnv.JAVA_HOME', os.environ.get('JAVA_HOME'))
--- a/SparkConnector/sparkconnector/configuration.py
+++ b/SparkConnector/sparkconnector/configuration.py
@@ -272,8 +272,7 @@ class SparkK8sConfiguration(SparkConfiguration):
         # The image used by the Spark executor should be set by the spawner, as we need
         # different images for different platform types
         # Get the value of 'SPARK_K8S_EXECUTOR_CONTAINER_IMAGE_URL' environment variable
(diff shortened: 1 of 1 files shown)


SparkConnector/sparkconnector/configuration.py
```diff
@@ -269,6 +269,15 @@ def configure(self, opts, ports):
         conf.set('spark.kubernetes.namespace', os.environ.get('SPARK_USER'))
         conf.set('spark.master', self._retrieve_k8s_master(os.environ.get('KUBECONFIG')))
 
+        # The image used by the Spark executor should be set by the spawner, as we need
+        # different images for different platform types
+        # Get the value of 'SPARK_K8S_EXECUTOR_CONTAINER_IMAGE_URL' environment variable
```
etejedor, line 274: Can we make the name of the variable more generic? Potentially we could use this for the physical Hadoop clusters too, right? Something like `SPARK_EXECUTOR_IMAGE`.

SparkConnector/sparkconnector/configuration.py
```diff
@@ -269,6 +269,15 @@ def configure(self, opts, ports):
...
+        # The default value is a stopgap till we implement the changes in the spawner
```
etejedor, line 275: This comment will get outdated soon, right? :) Similarly, the default could be made Alma9 already?

SparkConnector/sparkconnector/configuration.py
```diff
@@ -269,10 +269,13 @@ def configure(self, opts, ports):
         conf.set('spark.kubernetes.namespace', os.environ.get('SPARK_USER'))
         conf.set('spark.master', self._retrieve_k8s_master(os.environ.get('KUBECONFIG')))
 
-        # Configure shuffle if running on K8s with Spark 3.x.x
-        if self.get_spark_version().split('.')[0]=='3':
-            conf.set('spark.shuffle.service.enabled', 'false')
-            conf.set('spark.dynamicAllocation.shuffleTracking.enabled', 'true')
+        # The image used by the Spark executor should be set by the spawner, as we need
+        # different images for different platform types
```
diocas, line 273: Can't we check the platform directly in runtime and adapt then? (instead of having to set an extra config in the spawner, that might be out of sync)

diocas: Which was already there.

etejedor: What is already there exactly? We make available the tag as an environment variable (VERSION_DOCKER_IMAGE), but not the whole URL. Also, note that Luca uses his own images, so not even that tag is useful. In the current solution, the URLs are hardcoded in the SparkConnector's code, which should be avoided.

@LucaCanali @diocas we could think of just reusing the SWAN image for the Spark executors, get the tag from the environment (which is already there) and the rest of the URL from some configuration file for the SparkConnector -- what is dynamic is the tag, not the URL itself, but better in a config file than in the SparkConnector's code.

etejedor:
This is superseded by: https://github.com/swan-cern/jupyter-extensions/pull/372/files

Since we are migrating to Alma9 for Spark k8s.

The solution in the PR above is still hardcoding the URL of the image though. Ideally to be transformed into a configuration option for the connector.

PR #384, opened 2024-07-10
Changes on custom environments

RodrigoSobral2000:
These are some changes on how the extension approaches the arguments, including:

A new repo_type arguments comes from the frontend and is made for making the segregation on how the repostiory is treated and manipulated (can be either eos or git)
Verification on the structure of the repository is now done the extension, instead of the spawner, with the help of repo_type
On UI, distance between buttons and output box is now a bit bigger
Minor changes on comments and code details

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
@@ -77,7 +77,10 @@ done
 # --------------------------------------------------------------------------------------------
 # Validate input arguments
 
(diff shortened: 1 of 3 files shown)


etejedor:
This PR now adds on top of the initial implementation of the jupyter extension. This means that we need to separate changes into different commits.

Please split those changes into commits that contain changes that are related. One commit needs to contain changes that need to go as a whole, even across multiple files, or the repository would be left in an unconsistent state after the commit. Individual bug fixes or esthetic changes needs their own commits. Of course, each commit needs its commit message that explains the changes.

etejedor (approved):
Great, thanks!

PR #388, opened 2024-08-05
swancustomenvironments: change defaults, comments, and loggings

RodrigoSobral2000:
- Default for args is an empty string instead of `None`. This is done because bash scripts interpreter `None` as a normal string. Empty strings are the proper way to inform the system the variables are empty.
- Allow the git url pattern to accept `.git` extension
- Comments' explanation simplified
- Redirecting output also when creating the environments and symlinking the kernels (useful in case of errors)

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/serverextension.py
+++ b/SwanCustomEnvironments/swancustomenvironments/serverextension.py
@@ -22,16 +22,16 @@ class SwanCustomEnvironmentsApiHandler(APIHandler):
         """
         self.set_header("Content-Type", "text/event-stream")
 
-        repository = self.get_query_argument("repo", default=None)
-        repository_type = self.get_query_argument("repo_type", default=None)
-        accpy_version = self.get_query_argument("accpy", default=None)
+        repository = self.get_query_argument("repo", default="")
+        repository_type = self.get_query_argument("repo_type", default="")
+        accpy_version = self.get_query_argument("accpy", default="")
 
         arguments = ["--repo", repository, "--repo_type", repository_type]
-        if accpy_version is not None:
+        if accpy_version:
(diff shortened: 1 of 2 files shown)


etejedor:
Can you split this into as many commits as (possibly) the bullets you have in your commit message? Each with its own message.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -22,16 +22,16 @@ def get(self):
         """
         self.set_header("Content-Type", "text/event-stream")
 
-        repository = self.get_query_argument("repo", default=None)
-        repository_type = self.get_query_argument("repo_type", default=None)
-        accpy_version = self.get_query_argument("accpy", default=None)
+        repository = self.get_query_argument("repo", default='')
```
etejedor, line 25: Style comment: it is better if you stick to either single or double quotes for string literals throughout a document. Since you already have double quotes (at least in this fragment), I would use double for all the changes of this PR.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -22,16 +22,16 @@ def get(self):
...
+        repository_type = self.get_query_argument("repo_type", default='')
+        accpy_version = self.get_query_argument("accpy", default='')
 
         arguments = ["--repo", repository, "--repo_type", repository_type]
-        if accpy_version is not None:
+        if accpy_version != '':
```
etejedor, line 30: If you know `accpy_version` is always going to be a string, you can do:
```
if accpy_version:
```
instead to check if it's length is greater that zero.

etejedor (approved):

PR #389, opened 2024-08-05
swancustomenvironments: New conda environment option

RodrigoSobral2000:
Conda (mamba) environments can now be built passing the argument `--conda` into `makenv.sh` script. Some logic and code flow modifications were needed in order to build them sameway as `python-venv`.

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# Checks if the provided Acc-Py version is valid
+if [ ! -e "$ACCPY_PATH/base/$BUILDER_VERSION" ]; then
+    _error "Invalid Acc-Py version (${BUILDER_VERSION})."
+fi
+
+# Set up Acc-Py and create the environment
+source "${ACCPY_PATH}/base/${BUILDER_VERSION}/setup.sh"
+acc-py venv ${ENV_PATH} | tee -a ${LOG_PATH}
+
+# Activate the environment
+_log "Setting up the environment..."
+ACTIVATE_ENV_CMD="source ${ENV_PATH}/bin/activate"
+eval "${ACTIVATE_ENV_CMD}"
+
+# Install packages in the environment
+_log "Installing packages from ${REQ_PATH}..."
+pip install -r "${REQ_PATH}" | tee -a ${LOG_PATH}
+
+# Install the same ipykernel that the Jupyter server uses
+pip install ipykernel==${IPYKERNEL_VERSION} | tee -a ${LOG_FILE}
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/builders/mamba.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/builders/mamba.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Create the environment and install packages
+mamba create -p ${ENV_PATH} --file ${REQ_PATH} -y | tee -a ${LOG_PATH}
+
+# Activate the environment
+_log "Setting up the environment..."
+
+# Needed to use mamba activate without creating a new shell process
+MAMBADIR=$(mktemp -d)
+HOME=${MAMBADIR} mamba init
+ACTIVATE_MAMBA_CMD="source ${MAMBADIR}/.bashrc"
+eval "${ACTIVATE_MAMBA_CMD}"
+
+# Then activate the environment using mamba
+ACTIVATE_ENV_CMD="mamba activate ${ENV_PATH}"
+eval "${ACTIVATE_ENV_CMD}"
+
+# Install the same ipykernel that the Jupyter server uses
+mamba install "ipykernel==${IPYKERNEL_VERSION}" -y | tee -a ${LOG_PATH}
+
+# Source the mamba init script in the user's bash profile
+echo "${ACTIVATE_MAMBA_CMD}" >> /home/$USER/.bash_profile
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
@@ -33,11 +33,12 @@ define_repo_path() {
 
 # Function for printing the help page
 print_help() {
-    _log "Usage: makenv --repo/-r REPOSITORY [--repo_type TYPE] [--accpy ACCPY_VERSION] [--help/-h]"
+    _log "Usage: makenv --repo REPOSITORY --repo_type TYPE --builder BUILDER --builder_version VERSION [--help/-h]"
     _log "Options:"
-    _log "  -r, --repo REPOSITORY       Path or http link for a public repository (mandatory)"
-    _log "  --repo_type TYPE            Type of repository (git or eos) (mandatory)"
-    _log "  --accpy VERSION             Version of Acc-Py to be used"
+    _log "  --repo REPOSITORY           Path or http link for a public repository"
+    _log "  --repo_type TYPE            Type of repository (git or eos)"
+    _log "  --builder BUILDER           Builder to create the environment"
+    _log "  --builder_version VERSION   Version of the builder to use (optional)"
     _log "  -h, --help                  Print this help page"
 }
 
@@ -47,18 +48,47 @@ print_help() {
 while [ $# -gt 0 ]; do
     key="$1"
     case $key in
-        --repo|-r)
+        --repo)
             REPOSITORY=$2
+            # Check if a repository was provided
+            if [ -z "$REPOSITORY" ]; then
+                _log "ERROR: No repository provided." && log
+                print_help
+                exit 1
+            fi
             shift
             shift
             ;;
         --repo_type)
             REPO_TYPE=$2
+            # Check if a repository type was provided
+            if [ -z "$REPO_TYPE" ]; then
+                _log "ERROR: No repository type provided." && log
+                print_help
+                exit 1
+            fi
             shift
             shift
             ;;
-        --accpy)
-            ACCPY_VERSION=$2
+        --builder)
+            BUILDER=$2
+            BUILDER_PATH="$(dirname "$0")/builders/${BUILDER}.sh"
+            # Check if a builder was provided
+            if [ -z "$BUILDER" ]; then
+                _log "ERROR: No builder provided." && log
+                print_help
+                exit 1
+            # Check if is a valid builder
+            elif [ ! -f "${BUILDER_PATH}" ]; then
+                _log "ERROR: Invalid builder (${BUILDER})." && log
+                print_help
+                exit 1
+            fi
+            shift
+            shift
+            ;;
+        --builder_version)
+            BUILDER_VERSION=$2
             shift
             shift
             ;;
@@ -67,7 +97,7 @@ while [ $# -gt 0 ]; do
             exit 0
             ;;
         *)
-            _log "ERROR: Invalid argument: $1" && _log
+            _log "ERROR: Invalid argument: $1" && log
             print_help
             exit 1
             ;;
@@ -82,25 +112,8 @@ REPO_GIT_PATTERN='^https?:\/\/(github\.com|gitlab\.cern\.ch)\/([a-zA-Z0-9_-]+)\/
 # EOS path pattern: $CERNBOX_HOME/<folder1>/<folder2>/...(/?) or /eos/user/<lowercase_first_letter>/<username>/<folder1>?/<folder2>?/...(/?)
 REPO_EOS_PATTERN='^(\$CERNBOX_HOME(\/[^<>|\\:()&;,\/]+)*\/?|\/eos\/user\/[a-z](\/[^<>|\\:()&;,\/]+)+\/?)$'
 
-# Checks if the provided Acc-Py version is valid
-if [ -n "$ACCPY_VERSION" ] && [ ! -e "$ACCPY_PATH/base/$ACCPY_VERSION" ]; then
-    _log "ERROR: Invalid Acc-Py version (${ACCPY_VERSION})."
-    exit 1
-fi
-
-# Checks if a repository is provided
-if [ -z "$REPOSITORY" ]; then
-    _log "ERROR: No repository provided." && _log
-    print_help
-    exit 1
-
-elif [ -z "$REPO_TYPE" ]; then
-    _log "ERROR: No repository type provided." && _log
-    print_help
-    exit 1
-
 # Checks if the provided repository is a valid URL
-elif [[ "$REPO_TYPE" == "git" ]] && [[ "$REPOSITORY" =~ $REPO_GIT_PATTERN ]]; then
+if [[ "$REPO_TYPE" == "git" ]] && [[ "$REPOSITORY" =~ $REPO_GIT_PATTERN ]]; then
     # Extract the repository name
     repo_name=$(basename $REPOSITORY)
     repo_name=${repo_name%.*}
@@ -109,7 +122,7 @@ elif [[ "$REPO_TYPE" == "git" ]] && [[ "$REPOSITORY" =~ $REPO_GIT_PATTERN ]]; th
 
     # Clone the repository
     _log "Cloning the repository from ${REPOSITORY}..."
-    rm -rf "${REPO_PATH}" && git clone $REPOSITORY -q "${REPO_PATH}" || { _log "ERROR: Failed to clone repository"; exit 1; } | tee -a "${LOG_FILE}"
+    rm -rf "${REPO_PATH}" && git clone $REPOSITORY -q "${REPO_PATH}" || { _log "ERROR: Failed to clone repository"; exit 1; } | tee -a ${LOG_FILE}
     ENV_NAME="${repo_name}_env"
 
 # Checks if the provided local repository is an EOS path and actually exists
@@ -147,39 +160,25 @@ if [ ! -f "${REQ_PATH}" ]; then
     exit 1
 fi
 
+# Check if an environment already exists in the session, avoiding multiple environments
 CURRENT_ENV_NAME=$(find "/home/$USER" -type d -name "*_env" | head -n 1 | cut -d '/' -f4)
 CURRENT_REPO_PATH=$(tail -n 1 "/home/$USER/.bash_profile" | cut -d ' ' -f2)
-
-# Check if an environment already exists in the session
 if [ -n "${CURRENT_ENV_NAME}" ]; then
     _log "ENVIRONMENT_ALREADY_EXISTS:${CURRENT_ENV_NAME}"
     _log "REPO_PATH:${CURRENT_REPO_PATH#$HOME}"
     exit 1
 fi
 
-# Create environment (acc-py or generic)
-if [ -n "$ACCPY_VERSION" ]; then
-    source $ACCPY_PATH/base/${ACCPY_VERSION}/setup.sh
-    acc-py venv ${ENV_PATH} | tee -a "${LOG_FILE}"
+if [ -n "${BUILDER_VERSION}" ]; then
+    _log "Creating environment ${ENV_NAME} using ${BUILDER} (${BUILDER_VERSION})..."
 else
-    _log "Creating environment ${ENV_NAME} using Generic Python..."
-    python -m venv ${ENV_PATH} | tee -a "${LOG_FILE}"
+    _log "Creating environment ${ENV_NAME} using ${BUILDER}..."
 fi
-
-_log "ENV_NAME:${ENV_NAME}"
-
+source "${BUILDER_PATH}"
+ 
 # Make sure the Jupyter server finds the new environment kernel in /home/$USER/.local
 mkdir -p /home/$USER/.local/share/jupyter/kernels
-ln -f -s ${ENV_PATH}/share/jupyter/kernels/${ENV_NAME} /home/$USER/.local/share/jupyter/kernels/${ENV_NAME} | tee -a "${LOG_FILE}"
-
-# Activate the environment
-_log "Setting up the environment..."
-source ${ENV_PATH}/bin/activate
-
-# Install packages in the environment and the same ipykernel that the Jupyter server uses
-_log "Installing packages from ${REQ_PATH}..."
-pip install ipykernel==${IPYKERNEL_VERSION} | tee -a "${LOG_FILE}"
-pip install -r "${REQ_PATH}" | tee -a "${LOG_FILE}"
+ln -f -s ${ENV_PATH}/share/jupyter/kernels/${ENV_NAME} /home/$USER/.local/share/jupyter/kernels/${ENV_NAME} | tee -a ${LOG_FILE}
 
 if [[ ${REPO_TYPE} == "git" ]]; then
     # Move the repository from /tmp to the $CERNBOX_HOME/SWAN_projects folder
@@ -188,10 +187,11 @@ if [[ ${REPO_TYPE} == "git" ]]; then
     REPO_PATH="${GIT_HOME}/$(basename $REPO_PATH)"
 fi
 
+_log "ENV_NAME:${ENV_NAME}"
 _log "REPO_PATH:${REPO_PATH#$HOME}"
 
-# Install a Jupyter kernel for the environment
-python -m ipykernel install --name "${ENV_NAME}" --display-name "Python (${ENV_NAME})" --prefix "${ENV_PATH}" | tee -a "${LOG_FILE}"
+# Setting JUPYTER_PATH prevents ipykernel installation from complaining about non-found kernelspec
+JUPYTER_PATH=${ENV_PATH}/share/jupyter python -m ipykernel install --name "${ENV_NAME}" --display-name "Python (${ENV_NAME})" --prefix "${ENV_PATH}" | tee -a ${LOG_FILE}
 
 # Ensure the terminal loads the environment and cds into the repository path
-echo -e "source /home/$USER/${ENV_NAME}/bin/activate\ncd ${REPO_PATH}" > /home/$USER/.bash_profile
+echo -e "${ACTIVATE_ENV_CMD}\ncd ${REPO_PATH}" >> /home/$USER/.bash_profile
--- a/SwanCustomEnvironments/swancustomenvironments/serverextension.py
+++ b/SwanCustomEnvironments/swancustomenvironments/serverextension.py
@@ -18,24 +18,24 @@ class SwanCustomEnvironmentsApiHandler(APIHandler):
         """
         Gets the arguments from the query string and runs the makenv.sh script with them.
         repo (str): The git URL or absolute unix path to the repository.
-        accpy (str): The version of Accpy to be installed in the environment (optional).
+        repo_type (str): The type of repository (git or eos).
+        builder (str): The builder used for creating the environment.
+        builder_version (str): The version of the specified builder.
         """
         self.set_header("Content-Type", "text/event-stream")
 
         repository = self.get_query_argument("repo", default="")
-        repository_type = self.get_query_argument("repo_type", default="")
-        accpy_version = self.get_query_argument("accpy", default="")
+        repo_type = self.get_query_argument("repo_type", default="")
+        builder = self.get_query_argument("builder", default="")
+        builder_version = self.get_query_argument("builder_version", default="")
 
-        arguments = ["--repo", repository, "--repo_type", repository_type]
-        if accpy_version:
-            arguments.extend(["--accpy", accpy_version])
-
+        arguments = ["--repo", repository, "--repo_type", repo_type, "--builder", builder, "--builder_version", builder_version]
         makenv_process = subprocess.Popen([self.makenv_path, *arguments], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
 
         for line in iter(makenv_process.stdout.readline, b""):
             self.write(f"data: {line.decode('utf-8')}\n\n")
             self.flush()
-            
+
         self.finish("data: EOF\n\n")
 
 
--- a/SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
+++ b/SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
@@ -147,8 +147,9 @@ data-base-url="{{base_url | urlencode}}"
 
     const urlParams = new URLSearchParams(window.location.search);
     var repository = urlParams.get('repo') || "";
-    var repository_type = urlParams.get('repo_type') || "";
-    var accpy_version = urlParams.get('accpy') || "";
+    var repo_type = urlParams.get('repo_type') || "";
+    var builder = urlParams.get('builder') || "";
+    var builder_version = urlParams.get('builder_version') || "";
     var file = `/${urlParams.get('file') || ""}`;
 
     // Environment creation elements
@@ -165,7 +166,7 @@ data-base-url="{{base_url | urlencode}}"
     var modalJupyter = document.getElementById("modal-lab-button");
 
     // Create EventSource to listen for updates from the backend
-    var eventSource = new EventSource(`${base_url}api/customenvs?repo=${encodeURIComponent(repository)}&repo_type=${encodeURIComponent(repository_type)}&accpy=${encodeURIComponent(accpy_version)}`);
+    var eventSource = new EventSource(`${base_url}api/customenvs?repo=${encodeURIComponent(repository)}&repo_type=${encodeURIComponent(repo_type)}&builder=${encodeURIComponent(builder)}&builder_version=${encodeURIComponent(builder_version)}`);
 
     // Update UI every time a message is received
     eventSource.onmessage = (event) => {
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/builders/venv.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/builders/venv.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+# Create the environment
+python3 -m venv ${ENV_PATH} | tee -a ${LOG_PATH}
+
+# Activate the environment
+_log "Setting up the environment..."
+ACTIVATE_ENV_CMD="source ${ENV_PATH}/bin/activate"
+eval "${ACTIVATE_ENV_CMD}"
+
+# Install packages in the environment
+_log "Installing packages from ${REQ_PATH}..."
+pip install -r "${REQ_PATH}" | tee -a ${LOG_PATH}
+
+# Install the same ipykernel that the Jupyter server uses
+pip install ipykernel==${IPYKERNEL_VERSION} | tee -a ${LOG_FILE}


SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -33,11 +33,12 @@ define_repo_path() {
 
 # Function for printing the help page
 print_help() {
-    _log "Usage: makenv --repo/-r REPOSITORY [--repo_type TYPE] [--accpy ACCPY_VERSION] [--help/-h]"
+    _log "Usage: makenv --repo/-r REPOSITORY [--repo_type TYPE] [--accpy ACCPY_VERSION] [--conda] [--help/-h]"
```
etejedor, line 36: To tidy this up now, we could have an argument called `--builder`, and that can be `venv`, `accpy` or `conda`, venv can be the default.
Then, for the version, we could have `--builder_version`, to be able to set it for `accpy`. What do you think?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -171,16 +183,28 @@ _log "ENV_NAME:${ENV_NAME}"
 # Make sure the Jupyter server finds the new environment kernel in /home/$USER/.local
 mkdir -p /home/$USER/.local/share/jupyter/kernels
 ln -f -s ${ENV_PATH}/share/jupyter/kernels/${ENV_NAME} /home/$USER/.local/share/jupyter/kernels/${ENV_NAME} | tee -a "${LOG_FILE}"
+# Avoid an ipykernel installation warning message
+export JUPYTER_PATH=${ENV_PATH}/share/jupyter:$JUPYTER_PATH
```
etejedor, line 187: What is this? This is not conda related because it is done for all the env types.

RodrigoSobral2000: Indeed. I realised this line is able to avoid that boring warning message saying the configured kernelspec wasn't found in a given list of paths.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
RodrigoSobral2000: Yup, that's exactly the change that I mention in the other PR. But that involves making a considerable amount of changes. That's why I think would make more sense to implement in a separate PR.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: Ok, then this needs to go in another commit with its own message.

etejedor: Also, does the variable need to be exported?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: Why so many changes? I would assume this only requires changing the way you call the `makenv.sh` script, which only happens in one place in the code of the SwanCustomEnvs extension, right?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -157,10 +162,17 @@ if [ -n "${CURRENT_ENV_NAME}" ]; then
...
 
-# Create environment (acc-py or generic)
+# Create environment (acc-py, conda or generic as default)
 if [ -n "$ACCPY_VERSION" ]; then
-    source $ACCPY_PATH/base/${ACCPY_VERSION}/setup.sh
-    acc-py venv ${ENV_PATH} | tee -a "${LOG_FILE}"
+    source "${ACCPY_PATH}/base/${ACCPY_VERSION}/setup.sh"
+    acc-py venv ${ENV_PATH} | tee -a ${LOG_FILE}
+elif [ -n "$CONDA_ENV" ]; then
+    _log "Creating environment ${ENV_NAME} using Mamba Conda with requirements from (${REQ_PATH})..."
```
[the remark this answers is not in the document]
etejedor: I would also change the message here to `Creating conda environment ${ENV_NAME}..." to make it consistent with the message below.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -171,16 +183,28 @@ _log "ENV_NAME:${ENV_NAME}"
 # Make sure the Jupyter server finds the new environment kernel in /home/$USER/.local
 mkdir -p /home/$USER/.local/share/jupyter/kernels
 ln -f -s ${ENV_PATH}/share/jupyter/kernels/${ENV_NAME} /home/$USER/.local/share/jupyter/kernels/${ENV_NAME} | tee -a "${LOG_FILE}"
+# Avoid an ipykernel installation warning message
+export JUPYTER_PATH=${ENV_PATH}/share/jupyter:$JUPYTER_PATH
 
 # Activate the environment
 _log "Setting up the environment..."
-source ${ENV_PATH}/bin/activate
+if [ -n "$CONDA_ENV" ]; then
```
etejedor, line 191: I think this would look nicer without so many if else.
What I proposed when we talked yesterday is to have individual scripts that the main makenv.sh script sources.

For example, so far I have identified three actions that differ depending on the environment type: create, activate, install. You could factor out the code that is specific for each builder into the auxiliary scripts.

This means you could have auxiliary scripts of the form:
- create_conda.sh, activate_venv.sh, activate_accpy.sh
- activate_conda.sh, ...

Or, alternatively, just one big script for each that does all the steps:
- conda.sh, venv.sh, accpy.sh

And the main script would just do:
`source ${builder_name}.sh` or `source create_${builder_name}.sh` etc.

And the main script would:

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -171,16 +183,28 @@ _log "ENV_NAME:${ENV_NAME}"
...
+    ACTIVATE_ENV_CMD="source ${ENV_PATH}/bin/activate"
+    $(echo $ACTIVATE_ENV_CMD)
+
+    # Install packages in the environment
+    _log "Installing packages from ${REQ_PATH}..."
+    pip install -r "${REQ_PATH}" | tee -a ${LOG_FILE}
+fi
 
 # Install packages in the environment and the same ipykernel that the Jupyter server uses
 _log "Installing packages from ${REQ_PATH}..."
```
etejedor, line 204: This message should be deleted.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: Those scripts could be in a subdirectory called `builders`.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -157,10 +162,17 @@ if [ -n "${CURRENT_ENV_NAME}" ]; then
...
-# Create environment (acc-py or generic)
+# Create environment (acc-py, conda or generic as default)
 if [ -n "$ACCPY_VERSION" ]; then
-    source $ACCPY_PATH/base/${ACCPY_VERSION}/setup.sh
-    acc-py venv ${ENV_PATH} | tee -a "${LOG_FILE}"
+    source "${ACCPY_PATH}/base/${ACCPY_VERSION}/setup.sh"
+    acc-py venv ${ENV_PATH} | tee -a ${LOG_FILE}
+elif [ -n "$CONDA_ENV" ]; then
+    _log "Creating environment ${ENV_NAME} using Mamba Conda with requirements from (${REQ_PATH})..."
+    mamba env create -f ${REQ_PATH} -n ${ENV_NAME} -y | tee -a ${LOG_FILE}
```
etejedor, line 171: Could we defer the installation of the packages to later? Here we just create an empty environment and we add the packages later. This would make it symmetrical with the other builders.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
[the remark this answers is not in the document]
RodrigoSobral2000: It has not any message because the acc-py does that already:
```
[1;32m=> [0m Acc-Py base 2023.06.2 is now active within this shell.
INFO:acc_py_devtools:Creating virtual environment in /home/rhenriqu/std_python_env
Setting up the environment...
```

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
RodrigoSobral2000: I believe so, in order for the system to actually find the kernelspec for the environment. But I'll test without exporting it and see if everything is ok.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
RodrigoSobral2000: Wouldn't we lose some of the flexibility that a single script provide and create some redundancy? Since some of the actions are done for several cases

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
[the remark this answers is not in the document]
etejedor: Fair enough. But if you adopt the strategy of the auxiliary scripts, what I would do is just have one line:
```bash
_log "Creating environment ${ENV_NAME} using ${builder_name}..."
```
and after
```bash
source create_${builder_name}.sh
```

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: What is the line of this script exactly that needs this variable definition to be done? It would be good to have the definition of the variable right before the command that requires it.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: It depends on how much redundancy we are talking about. Modularity has a lot of value when structuring code: if you need to change something specific to e.g. coda, you know where to go. I don't think there's some much redundancy and, if necessary, we can factor out shared code in further auxiliary scripts.

In the end this is mimicking polymorphism. Would you like a code that does:
```
if object.class == Dog: object.run_function_dog()
else if object.class == Bear: object.run_function_bear()
...
```
or you like better
```
object.run_function() --> polimorphism will call one implementation or the other depending on runtime type of object
```
?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
RodrigoSobral2000: `python -m ipykernel install --name "${ENV_NAME}" --display-name "Python (${ENV_NAME})" --prefix "${ENV_PATH}"`
This is the line where ipykernel installs the environment kernel.
As you can see, the prefix is the path to the environment. And unlike python and accpy, conda environments are installed on `/opt/conda/envs` (something like this).

etejedor: Then I believe you could just do:
```
# JUPYTER_PATH is set to that the installation of ipykernel does not complain about non-found kernelspec
JUPYTER_PATH=${ENV_PATH}/share/jupyter python -m ipykernel install --name "${ENV_NAME}" --display-name "Python (${ENV_NAME})" --prefix "${ENV_PATH}"
```

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
etejedor, line 1: Nice! Now we can see what is common and what is different.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/python3.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
etejedor, line 1: Can you name this script `venv` instead of `python3`?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Create environment and install packages
+${BUILDER_VERSION} env create -f ${REQ_PATH} -n ${ENV_NAME} -y
```
etejedor, line 4: Does it make sense to create here empty and install packages later? To keep the same structure as the others? Or not really worth it?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# With a symlink on the user's home, no other customenvs can be created
```
etejedor, line 6: What is this symlink for? To detect that the conda environment already exists? How come we need it for conda and not for the other types?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Checks if the provided Acc-Py version is valid
+if [ -n "$BUILDER_VERSION" ] && [ ! -e "$ACCPY_PATH/base/$BUILDER_VERSION" ]; then
```
etejedor, line 4: If `$BUILDER_VERSION` is not defined we will continue running the script, we need to protect for that case and error out too.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
```diff
@@ -0,0 +1,19 @@
...
+    echo "ERROR: Invalid Acc-Py version (${BUILDER_VERSION})."
+    exit 1
+fi
+
+# Activate Acc-Py and create environment
```
etejedor, line 9:
```suggestion
# Setup Acc-Py and create environment
```

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -124,14 +150,12 @@ elif [[ "$REPO_TYPE" == "eos" ]] && [[ "$REPOSITORY" =~ $REPO_EOS_PATTERN ]]; th
     ENV_NAME="$(basename $REPO_PATH)_env"
 
     if [ ! -d "${REPO_PATH}" ]; then
-        _log "ERROR: Invalid ${REPO_TYPE} repository (${REPO_PATH})." && _log
-        exit 1
+        _error "Invalid ${REPO_TYPE} repository (${REPO_PATH})."
```
etejedor, line 153: The changes related to this new `_error` function are not related to conda, are they? They should be in a separate commit.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -143,55 +167,31 @@ IPYKERNEL_VERSION=$(python -c "import ipykernel; print(ipykernel.__version__)")
...
-_log "Setting up the environment..."
-source ${ENV_PATH}/bin/activate
-
-# Install packages in the environment and the same ipykernel that the Jupyter server uses
-_log "Installing packages from ${REQ_PATH}..."
-pip install ipykernel==${IPYKERNEL_VERSION} | tee -a "${LOG_FILE}"
-pip install -r "${REQ_PATH}" | tee -a "${LOG_FILE}"
+ln -f -s ${ENV_PATH}/share/jupyter/kernels/${ENV_NAME} /home/$USER/.local/share/jupyter/kernels/${ENV_NAME} | tee -a ${LOG_FILE}
 
+# Now the environment is set up, and as git repos are installed in /tmp, move it to the $Home/SWAN_projects folder
```
etejedor, line 180: I feel this comment is redundant since you already have a similar one inside the if?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -33,11 +52,12 @@ define_repo_path() {
...
 # Function for printing the help page
 print_help() {
-    _log "Usage: makenv --repo/-r REPOSITORY [--repo_type TYPE] [--accpy ACCPY_VERSION] [--help/-h]"
+    _log "Usage: makenv --repo_type TYPE --repo REPOSITORY --builder BUILDER --builder_version VERSION [--help/-h]"
     _log "Options:"
-    _log "  -r, --repo REPOSITORY       Path or http link for a public repository (mandatory)"
-    _log "  --repo_type TYPE            Type of repository (git or eos) (mandatory)"
-    _log "  --accpy VERSION             Version of Acc-Py to be used"
+    _log "  --repo_type TYPE            Type of repository (git or eos)"
+    _log "  --repo REPOSITORY           Path or http link for a public repository"
```
etejedor, line 58: Would it make sense to move the `--repo` line before `--repo_type` and not after?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
etejedor: Maybe not, never mind.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/python3.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Create environment
```
etejedor, line 3:
```suggestion
# Create the environment
```

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
```diff
@@ -0,0 +1,12 @@
...
+ln -f -s /opt/conda/envs/${ENV_NAME} ${ENV_PATH}
+ENV_PATH="/opt/conda/envs/${ENV_NAME}"
```
etejedor, line 8: Why is changing `ENV_PATH` necessary?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -8,11 +8,30 @@
 LOG_FILE=/tmp/makenv.log # File to keep a backlog of this script output
 GIT_HOME="$HOME/SWAN_projects" # Path where git repositories are stored
 
+# Check if an environment already exists in the session, avoiding multiple environments
+CURRENT_ENV_NAME=$(find "/home/$USER" -type d -name "*_env" | head -n 1 | cut -d '/' -f4)
+CURRENT_REPO_PATH=$(tail -n 1 "/home/$USER/.bash_profile" | cut -d ' ' -f2)
+
+# Check if an environment already exists in the session
```
etejedor, line 15: This comment is almost the same as the previous one.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -8,11 +8,30 @@
...
+if [ -n "${CURRENT_ENV_NAME}" ]; then
+    echo "ENVIRONMENT_ALREADY_EXISTS:${CURRENT_ENV_NAME}"
```
etejedor, line 17: Is this new check only necessary now with conda? If it's not related, please in its own commit with its message.

SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
```diff
@@ -146,9 +146,10 @@ <h2 id="modal-title" style="text-align: left;"></h2>
     var has_failed = false, is_created = false;
 
     const urlParams = new URLSearchParams(window.location.search);
+    var repo_type = urlParams.get('repo_type') || "";
```
etejedor, line 149: Same as above, makes sense to define repo type after repository?

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -17,25 +17,28 @@ class SwanCustomEnvironmentsApiHandler(APIHandler):
...
-        repository_type = self.get_query_argument("repo_type", default="")
-        accpy_version = self.get_query_argument("accpy", default="")
-
-        arguments = ["--repo", repository, "--repo_type", repository_type]
-        if accpy_version:
-            arguments.extend(["--accpy", accpy_version])
+        builder = self.get_query_argument("builder", default="")
+        builder_version = self.get_query_argument("builder_version", default="")
 
+        arguments = ["--repo_type", repo_type, "--repo", repository, "--builder", builder, "--builder_version", builder_version]
```
etejedor, line 32: Repo before repo_type?

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -17,25 +17,28 @@ class SwanCustomEnvironmentsApiHandler(APIHandler):
...
         Gets the arguments from the query string and runs the makenv.sh script with them.
+        repo_type (str): The type of repository (git or eos).
         repo (str): The git URL or absolute unix path to the repository.
-        accpy (str): The version of Accpy to be installed in the environment (optional).
+        builder (str): The builder used for creating the environment.
+        builder_version (str): The version of the specified builder.
         """
         self.set_header("Content-Type", "text/event-stream")
 
+        repo_type = self.get_query_argument("repo_type", default="")
```
etejedor, line 27: Repository before repo_type?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/python3.sh
RodrigoSobral2000: Yeah, but changing `python3-venv` to `venv-python3` is something that also has to be done on `options_from_form` (gitops)

etejedor: Ok, I think it makes sense though.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
RodrigoSobral2000: Conda has the following option
```
-p PATH, --prefix PATH
                        Full path to environment location (i.e. prefix).
```
However, when using it by itself, we get the following error:
```
mamba env create -p /home/rhenriqu/test_env 
...
EnvironmentFileNotFound: '/eos/home-i02/r/rhenriqu/environment.yml' file not found
```
It is complaining about not getting a requirements file. So, passing one, we get this new error:
```
mamba env create -p /home/rhenriqu/test_env -f local/std_python/requirements.txt 
...
SpecNotFound: Environment with requirements.txt file needs a name
```
This time, we need to provide a name for the environment
```
mamba env create -p /home/rhenriqu/test_env -f local/std_python/requirements.txt -n test_env
usage: mamba create [-h] [-f FILE] [-n ENVIRONMENT | -p PATH] [-C] [-k] [--offline] [--force] [--no-default-packages] [--subdir SUBDIR]
                    [--json] [-v] [-q] [-d] [-y] [--solver {classic,libmamba}]
                    [remote_definition]
mamba create: error: argument -n/--name: not allowed with argument -p/--prefix
```

So, the only option left is to create the environment only specifing name and requirements.
This is the reason why, as mentioned previously, conda environments are created under `/opt/conda/...`
The symlink is made for creating a reference under `/home/$user` so the script above is able to detect it and not allowing the creation of any other customenv

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
RodrigoSobral2000: Actually that initial condition is redundat. The `$BUILDER_VERSION ` is mandatory and the script stops at the beginning of `makenv.sh`

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
RodrigoSobral2000: I thought that if the `repository` value depends/is defined on/by `repo_type`,  would make more sense to prioritize `repo_type`

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
RodrigoSobral2000: ENV_PATH needs to be change due to the explanation above, the conda env is not under `/home/$user` anymore, so we need to point it to the new actual path

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
etejedor: For me this works:
```
mamba create -p /root/new_env --file /root/requirements.txt
```
Can you try with `--file` instead of `-f`?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
RodrigoSobral2000: This check isn't new. Is was just moved to the begining of the file to avoid running code unnecessarily. This is made for checking if there's any custom environment created already, no matter the name or type.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
RodrigoSobral2000: No way it worked.. 😮‍💨

RodrigoSobral2000: Changing...

RodrigoSobral2000: Ahh but hold on, the difference is not between `-f` and `--file`. You just used `mamba create` instead of `mamba env create` idk if there's any difference... Gotta do some research

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
etejedor: I believe after my suggestion for the command, the conda env can be under /home/$user, so you do not need to symlink nor update ENV_PATH.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: Ok so it's not conda related. It's own commit and message then, please.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: That would be fine if the second param was called `repo_location`, but being called `repo`... It's a bit like `builder` and `builder_version`.

etejedor (approved):

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -147,39 +166,23 @@ if [ ! -f "${REQ_PATH}" ]; then
...
-    acc-py venv ${ENV_PATH} | tee -a "${LOG_FILE}"
-else
-    _log "Creating environment ${ENV_NAME} using Generic Python..."
-    python -m venv ${ENV_PATH} | tee -a "${LOG_FILE}"
-fi
+_log "Creating environment ${ENV_NAME} using ${BUILDER} (${BUILDER_VERSION})..."
+source "${BUILDER_PATH}"
 
-_log "ENV_NAME:${ENV_NAME}"
+$(echo ${ACTIVATE_ENV_CMD})
```
etejedor, line 181: Is this echo a debug message that we needed to remove?

RodrigoSobral2000: the `ENV_NAME` log was moved down and that echo is the way I found for making sure it is activated no matter the command specified in the builder scripts

etejedor: Ah ok it's an execution, never mind.

RodrigoSobral2000: I know it may be redundant because is activated already in the builder scripts, its just a preventive measure. If not needed, it'll be removed in future PRs
The `ACTIVATE_ENV_CMD` is also useful for activating the environment everytime the terminal is opened in the lab.

etejedor: Ah true, why the need to activate again?

etejedor (dismissed):

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -188,10 +189,13 @@ if [[ ${REPO_TYPE} == "git" ]]; then
     REPO_PATH="${GIT_HOME}/$(basename $REPO_PATH)"
 fi
 
+_log "ENV_NAME:${ENV_NAME}"
 _log "REPO_PATH:${REPO_PATH#$HOME}"
 
-# Install a Jupyter kernel for the environment
-python -m ipykernel install --name "${ENV_NAME}" --display-name "Python (${ENV_NAME})" --prefix "${ENV_PATH}" | tee -a "${LOG_FILE}"
+# Install the same ipykernel that the Jupyter server uses
+pip install ipykernel==${IPYKERNEL_VERSION} | tee -a ${LOG_FILE}
```
etejedor, line 196: I would prefer to move the installation of ipykernel into the individual scripts and, in the case of conda-mamba, install ipykernel with mamba and not pip. Even if pip inside a conda environment can work, I think it's better that we use mamba to install packages that are available in coda repos. Here some people list reasons not to do pip inside conda:

https://www.reddit.com/r/learnpython/comments/q4u6ng/how_bad_is_installing_pip_packages_in_a_conda/

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# Create the environment and install packages
+${BUILDER_VERSION} create -p ${ENV_PATH} --file ${REQ_PATH} -y | tee -a ${LOG_PATH}
+
+# Activate the environment
+_log "Setting up the environment..."
+ACTIVATE_ENV_CMD="source activate ${ENV_PATH}"
```
etejedor, line 8: What did you find in the end about this?

From what I can read, `source activate` was the way to activate environments in older versions of conda. It is now not a documented way in the conda docs:

https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html#activating-an-environment

Did you find why our conda does not have the activate option? Is it just a matter of upgrading our conda in the user image? That could be done.

RodrigoSobral2000: From what I searched, I believe with these new changes it may work. I still gotta test it some more times but it looks ok from what I got. It could not find `conda activate` because we needed to `source /opt/conda/etc/profile.d/conda.sh` in order to properly recognize conda commands, kinda similar to `acc-py` in the sense that we first need to activate acc-py and then create the environment.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
RodrigoSobral2000: From the tests I did, I noticed conda has ipykernel already. Do you still want to keep it for consistency purposes or do you think it could be avoided to get the environment created faster?

etejedor: Since we enforce a certain ipykernel version, we need to install ipykernel explicitly.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
etejedor: What are the effects of running `source /opt/conda/etc/profile.d/conda.sh`? Does it have any implication for the environment?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Create the environment and install packages
+${BUILDER_VERSION} create -p ${ENV_PATH} --file ${REQ_PATH} -y | tee -a ${LOG_PATH}
+
+# Activate the conda to recognize all conda commands in the shell
+ACTIVATE_CONDA_SHELL="source /opt/conda/etc/profile.d/conda.sh"
+$(echo ${ACTIVATE_CONDA_SHELL}) | tee -a ${LOG_PATH}
```
etejedor, line 8: What's the advantage of running these commands with variable definition + echo? It just adds more lines, no?

RodrigoSobral2000: We use the exact same command (for activating the environment) when we're creating it and each time the terminal is opened (we want to activate it as well)

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
RodrigoSobral2000: It wouldn't have negative implications I believe, since all it does is defining conda related variables and functions. Meanwhile, yesterday evening, I found out that we can make it even simpler. Instead of sourcing `/opt/conda/etc/profile.d/conda.sh`, we can source `/home/$USER/.bashrc`. It changes a couple variables, including `$PATH`. I still need to fully understand why and how this is working.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
etejedor: Yes I see the same pattern is used in multiple places, but does it help in any way? Wouldn't we save a few code lines just doing source | tee, with a comment before that says what the source does?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
etejedor: What else is in .bashrc? Any possible side effect?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
RodrigoSobral2000: I thought tee commands was only used for input and output purposes. How would you run a command and write itself on a file for running it again later?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
RodrigoSobral2000: I believe not, this is what mine has
```
# .bashrc

# Source global definitions
if [ -f /etc/bashrc ]; then
        . /etc/bashrc
fi

# User specific environment
if ! [[ "$PATH" =~ "$HOME/.local/bin:$HOME/bin:" ]]
then
    PATH="$HOME/.local/bin:$HOME/bin:$PATH"
fi
export PATH

# Uncomment the following line if you don't like systemctl's auto-paging feature:
# export SYSTEMD_PAGER=

# User specific aliases and functions
if [ -d ~/.bashrc.d ]; then
        for rc in ~/.bashrc.d/*; do
                if [ -f "$rc" ]; then
                        . "$rc"
                fi
        done
fi

unset rc
eval "$(command conda shell.bash hook 2> /dev/null)"
```

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
RodrigoSobral2000: Done.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
RodrigoSobral2000: We got the scripts working fine. We're activating the environment using conda with no warning messages being displayed

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
etejedor: Doesn't
```
source /opt/conda/etc/profile.d/conda.sh | tee -a ${LOG_PATH}
```
achieve the same result as:
```
ACTIVATE_CONDA_SHELL="source /opt/conda/etc/profile.d/conda.sh"
$(echo ${ACTIVATE_CONDA_SHELL}) | tee -a ${LOG_PATH}
```
?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Check if the dependency is already in the requirements file, if not, add it to the file
```
etejedor, line 3: Why having this double block? Can't we just install our enforced version of ipykernel **after** installing the packages from the requirements file?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
etejedor: So iiuc the piece of code that enables conda commands is:
```
eval "$(command conda shell.bash hook 2> /dev/null)"
```
Why not just doing that before conda activate?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
RodrigoSobral2000: I initially tried that but you can really tell the difference between both approaches. I know it aint pretty but becomes much faster, so I went for that.

etejedor: Why is it faster? Because it does another round of resolving dependencies?
I think it is still better code-wise to be consistent with the other builders, where we first install requirements and later we install ipykernel with our fixed version.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/conda.sh
RodrigoSobral2000: Yup, but I was using it on makenv script too, that's why. But currently, we don't need to do that anymore

SwanCustomEnvironments/swancustomenvironments/scripts/builders/mamba.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# Create the environment and install packages
+mamba create -p ${ENV_PATH} --file ${REQ_PATH} -y | tee -a ${LOG_PATH}
+
+# Activate the environment
+_log "Setting up the environment..."
+
+# First source the .bashrc file so conda command can be properly recognized
```
etejedor, line 9: Change comment.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/mamba.sh
```diff
@@ -0,0 +1,17 @@
...
+eval "$(command conda shell.bash hook 2> /dev/null)"
```
etejedor, line 10:
```
# Needed to use mamba activate without creating a new shell process
source /opt/conda/etc/profile.d/mamba.sh
```

SwanCustomEnvironments/swancustomenvironments/scripts/builders/mamba.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Then activate the environment using conda
+ACTIVATE_ENV_CMD="conda activate ${ENV_PATH}"
```
etejedor, line 13: mamba activate

SwanCustomEnvironments/swancustomenvironments/scripts/builders/mamba.sh
```diff
@@ -0,0 +1,17 @@
...
+eval "${ACTIVATE_ENV_CMD}"
+
+echo "eval \"\$(command conda shell.bash hook 2> /dev/null)\"" >> /home/${USER}/.bash_profile
```
etejedor, line 16: This would not be needed.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/mamba.sh
```diff
@@ -0,0 +1,19 @@
...
+mamba create -p ${ENV_PATH} --file ${REQ_PATH} -y | tee -a ${LOG_PATH}
+
+# Activate the environment
+_log "Setting up the environment..."
+
+# Needed to use mamba activate without creating a new shell process
+HOME=/home/$USER mamba init
+ACTIVATE_MAMBA_CMD="source /home/$USER/.bashrc"
+eval "${ACTIVATE_MAMBA_CMD}"
+echo "${ACTIVATE_MAMBA_CMD}" >> /home/$USER/.bash_profile
```
etejedor, line 13: This means that, in `/home/$USER/.bash_profile`, we source both first the `.bashrc` in CERNBox (if it exists) and second the `.bashrc` in /home/$USER. I'm thinking whether it would be safer to source whatever the user has in CERNBox the last thing, to give priority to it.

Also, when we source /home/$USER/.bashrc, I noticed we also run:
```
eval "$(command conda shell.bash hook 2> /dev/null)"
```
And that activates a conda base environment, right? Is that what we want?

RodrigoSobral2000: 1:
`.bash_profile` comes with these lines by default. You're saying the priority order should be first `/home/$USER/.bashrc` and then `/eos./../.bashrc`?

RodrigoSobral2000: I don't think so. I listed environment packages afterwards and they were there looking good.

etejedor (approved):

PR #395, opened 2024-09-10
swancustomenvironments: Fixes and optimizations

RodrigoSobral2000:
Fix for error when calling extension (`/customenvironments`)
EventSource does not allow to pass headers. For that reason, a `fetch` promise is used. Its approach is different, however it has the same effect for this manner and no impact from the user's experience.
A dedidacted function for errors was added.
API code was adapted to reply accordingly.

---

Specify HOME to be users home instead of EOS for performance optimization when creating environment (including mamba)

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
@@ -174,7 +174,9 @@ if [ -n "${BUILDER_VERSION}" ]; then
 else
     _log "Creating environment ${ENV_NAME} using ${BUILDER}..."
 fi
-source "${BUILDER_PATH}"
+# To prevent builders (e.g. mamba) from caching files on EOS, which slows down the creation of the environment,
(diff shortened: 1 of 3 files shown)


SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -174,7 +174,8 @@ if [ -n "${BUILDER_VERSION}" ]; then
 else
     _log "Creating environment ${ENV_NAME} using ${BUILDER}..."
 fi
-source "${BUILDER_PATH}"
+# To avoid accessing EOS and bigger runtimes, configure HOME to be the user's home directory
```
etejedor, line 177: I would change this to:
```
# To prevent builders (e.g. mamba) from caching files on EOS, which slows down the creation of the environment,
# configure HOME to be the user's local directory
```

diocas (approved):

etejedor (approved):

PR #396, opened 2024-09-11
swancustomenvironments: Fix line breaks in output and piping to log file

RodrigoSobral2000:
- Insert line breaks between lines instead of chunks (some lines can be empty).
- Fix mistyped variables name from `LOG_PATH` to `LOG_FILE`

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
+++ b/SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
@@ -235,8 +235,9 @@ data-base-url="{{base_url | urlencode}}"
                 }
 
                 // Decode the chunk into text and split by newlines
-                const chunked_lines = decoder.decode(value, { stream: true });
-                chunked_lines.split('\n').forEach(message_line => {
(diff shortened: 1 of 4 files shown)


etejedor:
Quick comment: this needs two commits.

SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
```diff
@@ -237,31 +237,33 @@ <h2 id="modal-title" style="text-align: left;"></h2>
...
-                        is_created = true;
-                        env_name = message_line.split(':')[1];
-                    } 
-                    // Usual case -> update output box with the message received
-                    else {
-                        outputBox.innerHTML += message_line;
-                        outputBox.scrollTop = outputBox.scrollHeight;
-                        if (message_line.includes('ERROR')) {
-                            has_failed = true;
+                    if (message_line !== '') {
```
etejedor, line 240: If a line just contains a line break, it arrives here as an empty string I guess? Aren't with this change omitting the empty lines, and therefore they won't appear in the output?

PR #398, opened 2024-09-16
swancustomenvironments: Fix chunked_lines method

rodrigo-sobral:
Fix method for removing last element from list

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
+++ b/SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
@@ -236,7 +236,7 @@ data-base-url="{{base_url | urlencode}}"
 
                 // Decode the chunk into text and split by newlines
                 // Remove the last element (empty string) for avoiding an extra line break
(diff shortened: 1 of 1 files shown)


SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
```diff
@@ -236,7 +236,7 @@ <h2 id="modal-title" style="text-align: left;"></h2>
 
                 // Decode the chunk into text and split by newlines
                 // Remove the last element (empty string) for avoiding an extra line break
-                const chunked_lines = decoder.decode(value, { stream: true }).split('\n').pop();
+                const chunked_lines = decoder.decode(value, { stream: true }).split('\n').slice(0, -1);
```
etejedor, line 239:
```suggestion
                const chunked_lines = decoder.decode(value, { stream: true }).split('\n')[:-1];
```

rodrigo-sobral: I believe that doesnt work for Javascript

etejedor (approved):

PR #403, opened 2024-10-02
swancustomenvironments: Remove countdown

rodrigo-sobral:
Countdown is no longer needed, after meeting with Philip


etejedor:
So iiuc after these changes:
- When there is an error creating the environment, we do not go to JupyterLab, and instead we are shown the logs and a button appears to restart the session.
- When there is no error, we can see the logs as they are generated but when the environment is ready we go automatically to JupyterLab, and no button can be seen.

Is this correct?

rodrigo-sobral:
Yes, that's exactly what we want to achieve with this change.

etejedor (approved):

PR #408, opened 2024-11-06
sparkconnector: Preserve the default value of `extraJavaOptions`

rodrigo-sobral:
Customenv sessions don't have the same configuration as LCG software stack. Which leads to the lost of the spark defaults provided by NXCals. This specially affetcs extraJavaOptions, since its default is being overwritten by SparkConnector, but necessary to preserve, once it contains the list of NXCals instances for running jobs on.

### DIFF
--- a/SparkConnector/setup.py
+++ b/SparkConnector/setup.py
@@ -71,8 +71,7 @@ setup_args = dict(
     cmdclass= cmdclass,
     packages=setuptools.find_packages(),
     install_requires=[
-        "jupyterlab>=4.0.0,<5",
-        "bs4",
+        # "jupyterlab>=4.0.0,<5",
         "swanportallocator",
     ],
     zip_safe=False,
--- a/SparkConnector/sparkconnector/configuration.py
+++ b/SparkConnector/sparkconnector/configuration.py
@@ -1,5 +1,5 @@
 import os, subprocess, shutil, sys, uuid, time, base64, tempfile
-import requests
+import requests, re
 
 from pyspark import SparkConf, SparkContext
 from string import Formatter
@@ -132,11 +132,24 @@ class SparkConfiguration(object):
 
         # Extend conf adding logging of log4j to java options
         base_extra_java_options = "-Dlog4j.configuration=file:%s" % self.connector.log4j_file
-        extra_java_options = conf.get("spark.driver.extraJavaOptions")
-        if extra_java_options:
-            extra_java_options = base_extra_java_options + " " + extra_java_options
+        extra_java_options = " "
+
+        # For customenv sessions, load the default setting for spark.driver.extraJavaOptions from the
+        # spark-defaults.conf file in the environment, if it exists. Later, append the base_java_options.
+        # This is done to preserve the default setting and prevent base_java_options from completely
+        # replacing it. This is necessary e.g. for NXCALS node configuration.
+        if os.environ.get("SOFTWARE_SOURCE") == "customenv":
+            SPARK_DEFAULTS = os.environ.get(
+                "SPARK_DEFAULTS", 
+                os.path.join(os.environ.get("SPARK_HOME"), "conf", "spark-defaults.conf")
+            )
+            if os.path.exists(SPARK_DEFAULTS):
+                for line in open(SPARK_DEFAULTS).readlines():
+                    if line.startswith("spark.driver.extraJavaOptions"):
+                        extra_java_options = line.replace("spark.driver.extraJavaOptions", "").strip()
         else:
-            extra_java_options = base_extra_java_options
+            extra_java_options = conf.get("spark.driver.extraJavaOptions")
+        extra_java_options = f"{extra_java_options} {base_extra_java_options}"
         conf.set("spark.driver.extraJavaOptions", extra_java_options)
 
         # Extend conf ensuring that LD_LIBRARY_PATH on executors is the same as on the driver
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# If using NXCALS, we need to install the Spark extensions and the nxcals package.
+if [ -n "${INSTALL_NXCALS}" ]; then
+    SPARKCONNECTOR="sparkconnector==$(python -c 'import sparkconnector; print(sparkconnector.__version__)')"
+    SPARKMONITOR="sparkmonitor==$(python -c 'import sparkmonitor; print(sparkmonitor.__version__)')"
+    NXCALS="nxcals"
+fi
+
 # Set up Acc-Py and create the environment
 source "${ACCPY_PATH}/base/${BUILDER_VERSION}/setup.sh"
 acc-py venv ${ENV_PATH} | tee -a ${LOG_FILE}
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
@@ -196,7 +196,7 @@ ENV_PATH="/home/$USER/${ENV_NAME}"
 REQ_PATH="${REPO_PATH}/requirements.txt"
 IPYKERNEL_VERSION=$(python -c "import ipykernel; print(ipykernel.__version__)")
 
-# Libraries need to be installed and not linked, due to their dependencies
+# If using NXCALS, we need to install the Spark extensions and the nxcals package.
 if [ -n "${INSTALL_NXCALS}" ]; then
     SPARKCONNECTOR="sparkconnector==$(python3 -c 'import sparkconnector; print(sparkconnector.__version__)')"
     SPARKMONITOR="sparkmonitor==$(python3 -c 'import sparkmonitor; print(sparkmonitor.__version__)')"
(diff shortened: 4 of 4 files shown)


SparkConnector/sparkconnector/configuration.py
```diff
@@ -1,5 +1,5 @@
 import os, subprocess, shutil, sys, uuid, time, base64, tempfile
```
etejedor, line 1: The message of commit sparkconnector: Preserve the default value of extraJavaOptions  has typos, please fix.

SparkConnector/sparkconnector/configuration.py
```diff
@@ -132,11 +132,27 @@ def configure(self, opts, ports):
 
         # Extend conf adding logging of log4j to java options
         base_extra_java_options = "-Dlog4j.configuration=file:%s" % self.connector.log4j_file
-        extra_java_options = conf.get("spark.driver.extraJavaOptions")
-        if extra_java_options:
-            extra_java_options = base_extra_java_options + " " + extra_java_options
+        extra_java_options = None
+
+        # For customenv sessions, do not allow extraJavaOptions to be overwritten by the extra_java_options,
```
etejedor, line 137: Suggestion:
```
        # For customenv sessions, load the default setting for spark.driver.extraJavaOptions from the
        # spark-defaults.conf file in the environment, if it exists. Later, append the base_java_options.
        # This is done to preserve the default setting and prevent base_java_options from completely
        # replacing it. This is necessary e.g. for NXCALS node configuration.
```

SparkConnector/sparkconnector/configuration.py
```diff
@@ -132,11 +132,27 @@ def configure(self, opts, ports):
...
+
+        # For customenv sessions, do not allow extraJavaOptions to be overwritten by the extra_java_options, 
+        # since the NXCals instances, coming as a default value by the SparkContext when extraJavaOptions is
+        # not specified, need to be preserved in the driver configuration.
+        if os.environ.get("SOFTWARE_SOURCE") == "customenv":
+            SPARK_DEFAULTS = os.environ.get(
+                "SPARK_DEFAULTS", 
+                os.path.join(os.environ.get("SPARK_HOME"), "conf", "spark-defaults.conf")
+            )
+            if os.path.exists(SPARK_DEFAULTS):
```
etejedor, line 145: We need an else here: if the file does not exist, we just set `extra_java_options = base_extra_java_options`, as it is done in the case of the LCG release below.

SparkConnector/sparkconnector/configuration.py
```diff
@@ -132,11 +132,27 @@ def configure(self, opts, ports):
...
-        extra_java_options = conf.get("spark.driver.extraJavaOptions")
-        if extra_java_options:
-            extra_java_options = base_extra_java_options + " " + extra_java_options
+        extra_java_options = None
+
+        # For customenv sessions, do not allow extraJavaOptions to be overwritten by the extra_java_options, 
+        # since the NXCals instances, coming as a default value by the SparkContext when extraJavaOptions is
+        # not specified, need to be preserved in the driver configuration.
+        if os.environ.get("SOFTWARE_SOURCE") == "customenv":
+            SPARK_DEFAULTS = os.environ.get(
```
etejedor, line 141: Here, first line inside the if, I would initialize `extra_java_options = ''`. See why below.

SparkConnector/sparkconnector/configuration.py
```diff
@@ -132,11 +132,27 @@ def configure(self, opts, ports):
...
+                "SPARK_DEFAULTS", 
+                os.path.join(os.environ.get("SPARK_HOME"), "conf", "spark-defaults.conf")
+            )
+            if os.path.exists(SPARK_DEFAULTS):
+                for line in open(SPARK_DEFAULTS).readlines():
+                    if line.startswith("spark.driver.extraJavaOptions"):
+                        service_url = re.search(r'-Dservice\.url=[^\s]+', line).group(0)
```
etejedor, line 148: This only matches -Dservice.url= . But what if NXCALS change tomorrow the value of extraJavaOptions to include more options? You would miss them. You need to take everything that is defined as spark.driver.extraJavaOptions in that line.

SparkConnector/sparkconnector/configuration.py
```diff
@@ -132,11 +132,27 @@ def configure(self, opts, ports):
...
+                        extra_java_options = service_url + " " + base_extra_java_options
```
etejedor, line 149: Here you can set `extra_java_options` to just the value that you got from the spark-defaults.conf. Then, outside of the if statement, you can append `extra_java_options += base_extra_java_options`. This works both if `extra_java_options` was an empty string (it was initialized to that, see my comment above) and also in the case it took a value from spark-defaults.conf.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -222,6 +222,23 @@ JUPYTER_PATH=${ENV_PATH}/share/jupyter python -m ipykernel install --name "${ENV
 # We modify the already existing Python3 kernel with the kernel.json of the environment
 ln -f -s ${ENV_PATH}/share/jupyter/kernels/${ENV_NAME}/kernel.json /home/$USER/.local/share/jupyter/kernels/python3/kernel.json | tee -a ${LOG_FILE}
 
+# Pass Spark configurations to the environment kernel so it can be used in the notebooks
```
etejedor, line 217: Suggestion for comment:
```
# For NXCALS, configure the environment kernel and terminal with some variables to
# ensure the connection with the cluster works properly.
```

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -222,6 +222,23 @@ JUPYTER_PATH=${ENV_PATH}/share/jupyter python -m ipykernel install --name "${ENV
...
+if [ -n "${INSTALL_NXCALS}" ]; then
+    export SPARK_HOME="${ENV_PATH}/nxcals-bundle"
+    export PYSPARK_PYTHON="./environment/bin/python"
+    echo -e "export SPARK_HOME=\"${SPARK_HOME}\"\nexport PYSPARK_PYTHON=\"${PYSPARK_PYTHON}\"" >> /home/$USER/.bash_profile
```
etejedor, line 221: Please create two code blocks here, first one for the kernel and another one for the terminal, with a comment at the beginning of each block, e.g.

# Kernel configuration
# - SPARK_HOME: needed to point to the SPARK installation provided by the nxcals package
# - PYSPARK_PYTHON: ...
...

... 

# Terminal configuration
# Only SPARK_HOME and PYSPARK_PYTHON are needed, since `PATH` and `VIRTUAL_ENV` are already
# set when activating the environment in the terminal.  (I THINK THIS IS CORRECT, DOUBLE CHECK!)

...

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -222,6 +222,23 @@ JUPYTER_PATH=${ENV_PATH}/share/jupyter python -m ipykernel install --name "${ENV
...
+
+    # Passing configs to the kernel.json
+    export KERNEL_JSON=${ENV_PATH}/share/jupyter/kernels/${ENV_NAME}/kernel.json
```
etejedor, line 224: You don't need to export this variable, do you?

etejedor: See my comment above, to relocate this variable definition.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -222,6 +222,23 @@ JUPYTER_PATH=${ENV_PATH}/share/jupyter python -m ipykernel install --name "${ENV
 # We modify the already existing Python3 kernel with the kernel.json of the environment
 ln -f -s ${ENV_PATH}/share/jupyter/kernels/${ENV_NAME}/kernel.json /home/$USER/.local/share/jupyter/kernels/python3/kernel.json | tee -a ${LOG_FILE}
```
etejedor, line 215: You could define KERNEL_JSON here before the symlink and reuse it below.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -222,6 +222,23 @@ JUPYTER_PATH=${ENV_PATH}/share/jupyter python -m ipykernel install --name "${ENV
...
+    echo -e "export SPARK_HOME=\"${SPARK_HOME}\"\nexport PYSPARK_PYTHON=\"${PYSPARK_PYTHON}\"" >> /home/$USER/.bash_profile
+
+    # Passing configs to the kernel.json
+    export KERNEL_JSON=${ENV_PATH}/share/jupyter/kernels/${ENV_NAME}/kernel.json
+    jq --arg SPARK_HOME "${SPARK_HOME}" \
+       --arg PYSPARK_PYTHON "${PYSPARK_PYTHON}" \
+       --arg PATH "${PATH}" \
+       --arg VIRTUAL_ENV "${ENV_PATH}" \
+       '. + {env: {$SPARK_HOME, $PYSPARK_PYTHON, $PATH, $VIRTUAL_ENV}}' \
+       ${KERNEL_JSON} > /tmp/temp_kernel.json
```
etejedor, line 230: Do you really need a temporary or you can just write directly into KERNEL_JSON?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# Libraries need to be installed and not linked, due to their dependencies
```
etejedor, line 3: Suggestion for comment:
```
# If using NXCALS, we need to install the Spark extensions and the nxcals package.
```

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
```diff
@@ -1,5 +1,12 @@
...
+if [ -n "${INSTALL_NXCALS}" ]; then
+    SPARKCONNECTOR="sparkconnector==$(python -c 'import sparkconnector; print(sparkconnector.__version__)')"
```
etejedor, line 5: I am now thinking that we install sparkconnector and monitor together with their dependencies, right? I remember this pulled some extra packages that maybe we didn't want, but if we install them with --no-deps then they crash. Is this correct? What were the missing packages when installing with --no-deps?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -67,8 +67,7 @@ print_help() {
     _log "  --repo_type TYPE            Type of repository (git or eos)"
     _log "  --builder BUILDER           Builder to create the environment"
     _log "  --builder_version VERSION   Version of the builder to use (optional)"
-    _log "  --nxcals                    Install NXCALS package and Spark extensions in the environment (optional)"
-    _log "  -h, --help                  Print this help page"
+    _log "  --nxcals                    Install Nxcals extensions in the environment (optional)"    _log "  -h, --help                  Print this help page"
```
etejedor, line 70: Why this commit? Wasn't it ok as it was? If --nxcals is provided, we install both NXCALS and the Spark extensions.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
rodrigo-sobral: No, we don't. Fixing it

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
rodrigo-sobral: I tried and the file becomes empty. I believe it may be reading/writing the file at the same time

etejedor: Ok, then create the temporary file with `mktemp`.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
rodrigo-sobral: When we install sparkmonitor and sparkconnector without their dependencies (--no-deps) they start crashing. Sparkconnector, which is the first to be called, starts complaining about swanportallocator module not being found and so on, package after package. We're talking about around 40 libraries where some of them are jupyter and jupyterlab related and others like argon2, Jinja2, jsonpointer, etc

etejedor: From private conversation: testing if the jupyterlab dependency of sparkconnector can be skipped -- jupyterlab is the dependency that pulls most of the packages.

SparkConnector/setup.py
```diff
@@ -71,7 +71,6 @@
     cmdclass= cmdclass,
     packages=setuptools.find_packages(),
     install_requires=[
-        "jupyterlab>=4.0.0,<5",
```
etejedor, line 74: can you still keep it commented out?

etejedor: @diocas, we are commenting out this dependency because when installing sparkconnector in a custom environment, it pulls jupyterlab and a lot of dependencies with it, but we do not need all that to be installed in the environment (we tried). Do you have an objection to this?

etejedor: Discuss privately: we might be able to get rid of bs4 too.

etejedor (approved):

PR #411, opened 2024-11-11
swancustomenvironments: Implementation of sparkconnector bypass

rodrigo-sobral:
SparkConnector cannot be updated by now. Installing it by git repo takes much longer. So we install `swanportallocator` separately and  the current SparkConnector package with its dependencies longside


etejedor (approved):

etejedor:
As said in the commit message, this is to be changed when we create a new tag for the SparkConnector that we use in production for both LCG and custom environments.

PR #412, opened 2024-11-12
SwanCustomEnvironments, SparkConnector: Rebase on master, define SPARK_DRIVER_EXTRA_JAVA_OPTIONS

rodrigo-sobral:
SparkConnector cannot be updated by now. Installing it by git repo takes much longer. So we install `swanportallocator` separately and  the current SparkConnector package with its dependencies longside
Update SwanCustomEnvironments to v0.0.15
Define environment variable `SPARK_DRIVER_EXTRA_JAVA_OPTIONS` in the kernel, in order to SparkConnector configure extraJavaOptions with the default configuration


SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -189,14 +189,18 @@ if [ -n "${INSTALL_NXCALS}" ]; then
     # - PYSPARK_PYTHON: needed to point to the Python executable in the environment shipped to the cluster
     # - PATH: needed to point to all the binaries in the environment
     # - VIRTUAL_ENV: needed to point to the Python in the environment
+    # - SPARK_DRIVER_EXTRA_JAVA_OPTIONS: needed for gathering extraJavaOptions from the Spark default configuration, so it doesn't get overwritten by the SparkConnector
     export SPARK_HOME="${ENV_PATH}/nxcals-bundle"
     export PYSPARK_PYTHON="./environment/bin/python"
+    export SPARK_DRIVER_EXTRA_JAVA_OPTIONS=$(awk '/^spark.driver.extraJavaOptions/ {sub(/^spark.driver.extraJavaOptions /, ""); print}' ${SPARK_HOME}/conf/spark-defaults.conf)
```
etejedor, line 195: I believe we should place this variable definition in the acc-py.sh script, protected by an if so that it only gets done if NXCALS.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -189,14 +189,18 @@ if [ -n "${INSTALL_NXCALS}" ]; then
...
+
     NEW_KERNEL=$(mktemp)
     jq --arg SPARK_HOME "${SPARK_HOME}" \
     --arg PYSPARK_PYTHON "${PYSPARK_PYTHON}" \
     --arg PATH "${PATH}" \
     --arg VIRTUAL_ENV "${ENV_PATH}" \
-    '. + {env: {$SPARK_HOME, $PYSPARK_PYTHON, $PATH, $VIRTUAL_ENV}}' \
+    --arg SPARK_DRIVER_EXTRA_JAVA_OPTIONS "${SPARK_DRIVER_EXTRA_JAVA_OPTIONS}" \
```
etejedor, line 202: This will not work well if SPARK_DRIVER_EXTRA_JAVA_OPTIONS is not defined, right? Perhaps we should protect this variable inclusion with "if nxcals".

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: Clarified, ignore.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: Clarified, ignore.

etejedor (approved):

PR #413, opened 2024-11-12
sparkconnector: Use `SPARK_DRIVER_EXTRA_JAVA_OPTIONS`

rodrigo-sobral:
With `SPARK_DRIVER_EXTRA_JAVA_OPTIONS` defined as environment variable in the kernel, it configures `extraJavaOptions` with the default spark configuration.


SparkConnector/sparkconnector/configuration.py
```diff
@@ -132,11 +132,16 @@ def configure(self, opts, ports):
...
         base_extra_java_options = "-Dlog4j.configuration=file:%s" % self.connector.log4j_file
-        extra_java_options = conf.get("spark.driver.extraJavaOptions")
-        if extra_java_options:
-            extra_java_options = base_extra_java_options + " " + extra_java_options
+        extra_java_options = " "
+
+        # SPARK_DRIVER_EXTRA_JAVA_OPTIONS contains the default value of spark.driver.extraJavaOptions.
+        # Later, append the base_java_options. This is done to preserve the default setting and prevent 
+        # base_java_options from completely replacing it. This is necessary e.g. for NXCALS node configuration (customenvs).
+        if os.environ.get("SPARK_DRIVER_EXTRA_JAVA_OPTIONS"):
```
etejedor, line 140: I'm thinking about what logic we should follow here. We have three possible extra java options sources:
1. The variable
2. The user configuration via the Spark Connector UI
3. The logging option we need to add

I believe if 1 is defined, it should always be present because otherwise NXCALS can break. 3 also needs to be present always because otherwise the logging breaks. In the case of 2, it is options coming from the user. We could in this case prepend them to ours? So this means that if 1, 2 and 3 exist, the final options are : 2 + 1 + 3. Opinion?

SparkConnector/sparkconnector/configuration.py
```diff
@@ -132,11 +132,15 @@ def configure(self, opts, ports):
...
         # Extend conf adding logging of log4j to java options
         base_extra_java_options = "-Dlog4j.configuration=file:%s" % self.connector.log4j_file
-        extra_java_options = conf.get("spark.driver.extraJavaOptions")
-        if extra_java_options:
-            extra_java_options = base_extra_java_options + " " + extra_java_options
-        else:
-            extra_java_options = base_extra_java_options
+        extra_java_options = conf.get("spark.driver.extraJavaOptions", "")
+
+        # SPARK_DRIVER_EXTRA_JAVA_OPTIONS contains the default value of spark.driver.extraJavaOptions.
```
etejedor, line 137: I propose to replace this comment and use the one below for the whole block, starting with the first setting of extra_java_options:
```
# Fill extra_java_options as a concatenation of:
1. Options configured by the user via the UI of SparkConnector
2. SPARK_DRIVER_EXTRA_JAVA_OPTIONS variable, set for example to load required NXCALS settings
3. Base Java options with log4j configuration
```

etejedor: 1 user_extra_java_options
2 env_extra_java_options
3 logging_extra_java_options

PR #421, opened 2024-12-05
swancustomenvironments: Remove swan middle layer

rodrigo-sobral:
Due to the fact `venv-pack` (used by nxcals) doesn't support editable packages to be shipped with the environment, and the swan middle layer was based on the use of a .pth file, which is seen as a source of editable packages and therefore makes venv-pack fail. For this reason, the swan middle layer has been removed and the extra packages have been added directly to the environment. In spite of this, the requirements file must not be modified.

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
@@ -24,10 +24,16 @@ if [ "${RESOLVED_REQ}" = true ]; then
     # Use the same pip configuration as the Acc-Py default pip
     ACCPY_PIP_CONF="-i $(pip config get global.index-url) --allow-insecure-host $(pip config get global.trusted-host)"
     uv pip install ${ACCPY_PIP_CONF} -r "${REQ_PATH}" 2>&1
+    if [ $? -ne 0 ]; then
+        return 1
+    fi
     # Enforce installation of our version of ipykernel and its dependencies
     uv pip install ${ACCPY_PIP_CONF} ${IPYKERNEL} 2>&1
 else
     pip install -r "${REQ_PATH}" 2>&1
+    if [ $? -ne 0 ]; then
+        return 1
+    fi
     # Enforce installation of our version of ipykernel and its dependencies
     pip install ${IPYKERNEL} 2>&1
 fi
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
@@ -1,27 +1,10 @@
 #!/bin/bash
 
-# Create a middle layer for installing ipykernel, putting it apart from the user environment
-uv venv $SWAN_ENV --seed 2>&1
-source $SWAN_ENV/bin/activate
-uv pip install "ipykernel==${IPYKERNEL_VERSION}"
-SWAN_PACKAGES_PATH=$(python3 -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')
-deactivate
-
 # If using NXCALS, we need to also install the Spark packages and their dependencies in the SWAN environment
 if [ -n "${USE_NXCALS}" ]; then
     SPARKCONNECTOR="sparkconnector==$(python -c 'import sparkconnector; print(sparkconnector.__version__)')"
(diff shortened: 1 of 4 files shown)


SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
```diff
@@ -229,7 +229,7 @@ <h2 id="modal-title" style="text-align: left;"></h2>
                     else {
                         outputBox.innerHTML += message_line;
                         outputBox.scrollTop = outputBox.scrollHeight;
-                        if (message_line.includes('ERROR')) {
+                        if (message_line.toUpperCase().includes('ERROR')) {
```
etejedor, line 232: Why do you need the uppercase? Isn't ERROR in lowercase your signal for when something is wrong? Is it to catch other errors that you do  not signal?

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
```diff
@@ -1,27 +1,10 @@
 #!/bin/bash
 
-# Create a middle layer for installing ipykernel, putting it apart from the user environment
```
etejedor, line 3: The commit message still needs to reflect the fact that venv-pack does not support editable packages, and the middle layer was based on the use of a .pth file, which is seen as a source of editable packages and therefore makes venv-pack fail.

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
```diff
@@ -41,15 +24,28 @@ if [ "${RESOLVED_REQ}" = true ]; then
     # Use the same pip configuration as the Acc-Py default pip
     ACCPY_PIP_CONF="-i $(pip config get global.index-url) --allow-insecure-host $(pip config get global.trusted-host)"
     uv pip install ${ACCPY_PIP_CONF} -r "${REQ_PATH}" 2>&1
+    uv pip install ${ACCPY_PIP_CONF} ${IPYKERNEL} 2>&1
```
etejedor, line 30: Here and below, we need a comment that says:

```
# Enforce installation of our version of ipykernel
```

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
```diff
@@ -41,15 +24,28 @@ if [ "${RESOLVED_REQ}" = true ]; then
...
 if [ $? -ne 0 ]; then
     exit 1
 fi
 
-# Inject middle layer packages into the user environment by adding a .pth file to
-# the environment site-packages that contains the path to the middle layer site-packages
-USER_PACKAGES_PATH=$(python3 -c 'import sysconfig; print(sysconfig.get_paths()["purelib"])')
-echo ${SWAN_PACKAGES_PATH} > ${USER_PACKAGES_PATH}/$(basename $SWAN_ENV).pth
+if [ -n "${USE_NXCALS}" ]; then
+    # For NXCALS, install the Spark packages and their dependencies, using the according tool for its resolution
```
etejedor, line 40: Suggestion:
```
# For NXCALS, enforce installation of our Spark extensions and their dependencies at certain versions
```

SwanCustomEnvironments/swancustomenvironments/scripts/builders/venv.sh
```diff
@@ -24,14 +17,11 @@ eval "${ACTIVATE_ENV_CMD}"
 _log "Installing packages from ${REQ_PATH}..."
 if [ "${RESOLVED_REQ}" = true ]; then
     uv pip install -r "${REQ_PATH}" 2>&1
+    uv pip install ${IPYKERNEL} 2>&1
```
etejedor, line 23: Same comment as above

SwanCustomEnvironments/swancustomenvironments/scripts/builders/accpy.sh
```diff
@@ -24,14 +24,17 @@ if [ "${RESOLVED_REQ}" = true ]; then
     # Use the same pip configuration as the Acc-Py default pip
     ACCPY_PIP_CONF="-i $(pip config get global.index-url) --allow-insecure-host $(pip config get global.trusted-host)"
     uv pip install ${ACCPY_PIP_CONF} -r "${REQ_PATH}" 2>&1
+    if [ $? -ne 0 ]; then
+        return 1
```
etejedor, line 28: Why is it return 1 and not exit 1? exit 1 will make the source command exit with a value different than zero, no?

etejedor (approved):

PR #422, opened 2024-12-10
sparkconnector: Update dependencies

rodrigo-sobral:
Remove `bs4` and  `jupyterlab` and add  `requests` dependency

### DIFF
--- a/SparkConnector/setup.py
+++ b/SparkConnector/setup.py
@@ -71,7 +71,6 @@ setup_args = dict(
     cmdclass= cmdclass,
     packages=setuptools.find_packages(),
     install_requires=[
-        "jupyterlab>=4.0.0,<5",
         "swanportallocator",
     ],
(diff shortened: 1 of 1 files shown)


SparkConnector/setup.py
```diff
@@ -71,7 +71,6 @@
     cmdclass= cmdclass,
     packages=setuptools.find_packages(),
     install_requires=[
-        "jupyterlab>=4.0.0,<5",
```
etejedor, line 74: Well, it is a runtime dependency because the extension runs inside jupyterlab, right? I would say as commit message:
```
Even though SparkConnector is a lab extension, we can assume lab will be installed
when using the extension. Removing the explicit dependency helps us not pull
lab and all its dependencies when constructing a custom environment in SWAN.
```

etejedor (approved):

PR #424, opened 2025-02-13
Swan on tn multitabs

rodrigo-sobral:
With these changes, the extension is able to deal with multiple simultaneous requests, broadcasting its messages to all of them in real time. This is useful in order to avoid that a request gets blocked waiting for the first request to finish.

It also removes the "environment already created" blocking statement and modal

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
+++ b/SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
@@ -7,10 +7,7 @@
 
 # Check if an environment already exists in the session, avoiding multiple environments
 CURRENT_ENV_NAME=$(find "/home/$USER" -type d -name "*_env" | head -n 1 | cut -d '/' -f4)
-CURRENT_REPO_PATH=$(tail -n 1 "/home/$USER/.bash_profile" | cut -d ' ' -f2)
 if [ -n "${CURRENT_ENV_NAME}" ]; then
-    echo "ENVIRONMENT_ALREADY_EXISTS:${CURRENT_ENV_NAME}"
-    echo "REPO_PATH:${CURRENT_REPO_PATH#$HOME}"
     exit 1
 fi
 
--- a/SwanCustomEnvironments/swancustomenvironments/serverextension.py
+++ b/SwanCustomEnvironments/swancustomenvironments/serverextension.py
@@ -1,4 +1,5 @@
 from tornado import web
+import asyncio
 
 from jupyter_server.base.handlers import JupyterHandler, APIHandler
 from jupyter_server.utils import url_path_join
@@ -16,8 +17,10 @@ class SwanCustomEnvironmentsApiHandler(APIHandler):
     # Path to the file where the log of the makenv.sh script is written
     LOG_FILE = "/tmp/makenv.log"
 
+    makenv_process = None
+
     @web.authenticated
-    def get(self):
+    async def get(self):
         """
         Gets the arguments from the query string and runs the makenv.sh script with them.
         repo (str): The git URL or absolute unix path to the repository.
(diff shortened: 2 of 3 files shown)


etejedor:
Can you put everything in one commit? All these changes work together, so they have to be in a single unit of consistency.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -26,7 +29,28 @@ def get(self):
         nxcals (bool): Whether to include NXCALS and Spark extensions in the environment.
         """
         self.set_header("Content-Type", "text/event-stream")
+        running_process = SwanCustomEnvironmentsApiHandler.makenv_process
```
etejedor, line 32: Nice refactoring!

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -26,7 +29,28 @@ def get(self):
...
+        
+        log_file = None
+        if running_process is None:
```
etejedor, line 35: Can you call this variable `makenv_process`? I think it is more telling about what the process does.

rodrigo-sobral: Don't you think it can become too ambiguous?

etejedor, line 35: Inside this if we need a comment:
```
# The first get request will launch the execution of the makenv.sh script,
# to create the environment
```

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -4,16 +4,6 @@
 # Copyright CERN
 # This script allows to create an environment to use in notebooks and terminals. The environment contains the packages from a provided repository.
 
-
-# Check if an environment already exists in the session, avoiding multiple environments
```
diocas, line 8: While it makes some sense to stop checking in the caller/backend side (because we basically impose a singleton), I would argue this script should still verify and exit on error. Unless is ok to call makenv.sh multiple times, this should be protected as there could be multiple ways of calling it.

SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
```diff
@@ -159,12 +103,6 @@ <h2 id="modal-title" style="text-align: left;"></h2>
     var loader = document.getElementById('swan-loader');
     var back_button = document.getElementById('back-button');
 
-    // Modal elements
```
diocas, line 162: Would it be possible somehow to land in the extension with a diff configuration that the previously used? I would imagine we protect about any of this from the hub extension, right?

diocas (commented):
Nice change! Just a small request.

SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
etejedor: Yes we do protect. If the user requested a different configuration than the one that is being created / has been created, that configuration is discarded since they agreed in the dialog to proceed to the current session. So the configuration of the current session dominates and stays. If they want to use the second configuration, they should pick the option to terminate the current session in the dialog.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
etejedor: We actually were debating about this with @rodrigo-sobral , we see pros and cons in leaving / removing the code. Leaving it means more code to maintain that is actually not used in our workflow. Removing it removes the protection in case someone finds the script and invokes it manually, as you pointed out. I'm fine with leaving the code.

diocas: I would suggest leaving it then.

PR #425, opened 2025-02-20
swancustomenvironments: Multitab error handling

rodrigo-sobral:
With these changes, the extension is able to deal with errors when creating the environment in multiple simultaneous requests, sharing the same status with all tabs.

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/serverextension.py
+++ b/SwanCustomEnvironments/swancustomenvironments/serverextension.py
@@ -38,6 +38,15 @@ class SwanCustomEnvironmentsApiHandler(APIHandler):
             makenv_process = self._launch_makenv()
             SwanCustomEnvironmentsApiHandler.makenv_process = makenv_process
             log_file = open(self.LOG_FILE, "w")
+        else:
+            ret_code = makenv_process.poll()
+            if ret_code is not None and ret_code != 0:
+                # If the process has finished with error, show the content of the log file
+                with open(self.LOG_FILE, "r") as log_file:
+                    for line in log_file:
+                        await self._send_line(line)
+                self.finish()
+                return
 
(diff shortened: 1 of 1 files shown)


SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -38,6 +38,15 @@ async def get(self):
             makenv_process = self._launch_makenv()
             SwanCustomEnvironmentsApiHandler.makenv_process = makenv_process
             log_file = open(self.LOG_FILE, "w")
+        else:
```
etejedor, line 41: Perhaps a better way to check if the process failed is via its return code:
```python
else if makenv_process.returncode != 0:

if process.returncode != 0:
    with open(self.LOG_FILE, "r") as log_file:
        for line in log_file:
            self.write(line)
            await self.flush()
    self.finish()
    return
```

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -38,18 +38,21 @@ async def get(self):
             makenv_process = self._launch_makenv()
             SwanCustomEnvironmentsApiHandler.makenv_process = makenv_process
             log_file = open(self.LOG_FILE, "w")
+        elif makenv_process.poll() == 1:
+            # If the process has finished, open the log file in read mode to check for errors
```
etejedor, line 42: An error return code could also be different than 1 but still indicate some error. It has to be different than zero and different than None.

etejedor, line 42:
```suggestion
            # If the process has finished with error, show the content of the log file
```

etejedor (approved):

PR #427, opened 2025-02-26
swancontents: Use async file contents manager

PMax5:
Use an async file contents manager to have async file operations, such as for listing and saving files. This was implemented to solve an issue that was being caused by users not being able to type in terminal when they would open a folder with a large amount of files in the file browser of JupyterLab

By separating the file operations in different threads and releasing control of the main one, the issue has been fixed

### DIFF
--- a/SwanContents/swancontents/checkpoints/eoscheckpoints.py
+++ b/SwanContents/swancontents/checkpoints/eoscheckpoints.py
@@ -1,13 +1,14 @@
-from jupyter_server.services.contents.checkpoints import Checkpoints
+from jupyter_server.services.contents.checkpoints import AsyncCheckpoints
 from ..filemanager.eos.fileio import SwanFileManagerMixin
 import os
 import datetime
+from anyio.to_thread import run_sync
 from traitlets import Unicode, Int
 from tornado.web import HTTPError
 
 
 
-class EOSCheckpoints(SwanFileManagerMixin, Checkpoints):
+class EOSCheckpoints(SwanFileManagerMixin, AsyncCheckpoints):
     """
         Implements Checkpoints interface for EOS.
         This allows the creation of EOS versions, visible both from
@@ -37,7 +38,7 @@ class EOSCheckpoints(SwanFileManagerMixin, Checkpoints):
     latest_recorded = {}
 
 
-    def create_checkpoint(self, contents_mgr, path):
+    async def create_checkpoint(self, contents_mgr, path):
         """
             EOS creates automatically a version by using the atomic writing.
             So we just need to return the latest checkpoint created
@@ -45,7 +46,7 @@ class EOSCheckpoints(SwanFileManagerMixin, Checkpoints):
         self.log.info(f"Creating checkpoint for {path}")
         # To check if the version returned is new or already known (an error might have occurred)
         previous_recorded = self.latest_recorded[path] if path in self.latest_recorded else None
-        checkpoints = self.list_checkpoints(path)
+        checkpoints = await self.list_checkpoints(path)
 
         if not checkpoints:
             self.log.warning("No checkpoint was created")
@@ -62,11 +63,11 @@ class EOSCheckpoints(SwanFileManagerMixin, Checkpoints):
 
         return current_checkpoint
 
-    def restore_checkpoint(self, contents_mgr, checkpoint_id, path):
+    async def restore_checkpoint(self, contents_mgr, checkpoint_id, path):
         """ Replace the current file with a previous version"""
         checkpoint = self._get_checkpoint_info(path, checkpoint_id)
         try:
-            self._copy(checkpoint['checkpoint_path'], checkpoint['src_path'])
+            await self._copy(checkpoint['checkpoint_path'], checkpoint['src_path'])
         except: 
             # the version might no longer exist if it was cleaned (by default EOS should keep 10 versions)
             self._no_such_checkpoint(path, checkpoint_id)
@@ -80,7 +81,7 @@ class EOSCheckpoints(SwanFileManagerMixin, Checkpoints):
         """
         pass
 
-    def delete_checkpoint(self, checkpoint_id, path):
+    async def delete_checkpoint(self, checkpoint_id, path):
         """Remove a created version"""
         cp_path = self._get_checkpoint_info(path, checkpoint_id)['checkpoint_path']
 
@@ -89,14 +90,14 @@ class EOSCheckpoints(SwanFileManagerMixin, Checkpoints):
 
         self.log.debug("Unlinking checkpoint %s", cp_path)
         with self.perm_to_403():
-            os.unlink(cp_path)
+            await run_sync(os.unlink, cp_path)
 
-    def list_checkpoints(self, path):
+    async def list_checkpoints(self, path):
         """ On notebook opening, returns a list of all available versions. """
         base = self._get_checkpoint_base(path)
 
         try:
-            files = os.listdir(base['base_path'])
+            files = await run_sync(os.listdir, base['base_path'])
             files.sort()
             to_return = [self._get_checkpoint_return(file) for file in files]
             # Keep track of the latest version to compare when creating a new one
--- a/SwanContents/swancontents/filemanager/eos/fileio.py
+++ b/SwanContents/swancontents/filemanager/eos/fileio.py
@@ -1,7 +1,7 @@
-from jupyter_server.services.contents.fileio import FileManagerMixin
+from jupyter_server.services.contents.fileio import AsyncFileManagerMixin
 from jupyter_server.utils import url_path_join
 from tornado.web import HTTPError
-from contextlib import contextmanager
+from contextlib import contextmanager, asynccontextmanager
 import io, os, nbformat
 import subprocess
 
@@ -82,7 +82,7 @@ def atomic_writing(path, text=True, encoding='utf-8', log=None, **kwargs):
             subprocess.run(["setfattr","-x", "user.fusex.rename.version", dirname])
 
 
-class SwanFileManagerMixin(FileManagerMixin):
+class SwanFileManagerMixin(AsyncFileManagerMixin):
     """
     Mixin for ContentsAPI classes that interact with the filesystem.
 
--- a/SwanContents/swancontents/filemanager/eos/fileio.py
+++ b/SwanContents/swancontents/filemanager/eos/fileio.py
@@ -107,58 +107,6 @@ class SwanFileManagerMixin(FileManagerMixin):
         else:
             return super()._get_os_path(path)
 
-    def _save_notebook(self, os_path, nb, capture_validation_error=None):
-        """
-        Save a notebook on EOS via FUSE with the default routine.
-        Plus, store a copy of the same notebook on the host machine via docker volume.
-        """
-        import time
-        def write_notebook_to_local(path, content, encoding='utf-8'):
-            try:
-                fout = io.open(path, 'w', encoding=encoding)
-                nbformat.write(content, fout, version=nbformat.NO_CONVERT)
-                fout.flush()
-                os.fsync(fout.fileno())
-                fout.close()
-            except:
-                pass
-            return
-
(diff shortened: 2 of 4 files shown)


SwanContents/swancontents/checkpoints/eoscheckpoints.py
```diff
@@ -91,12 +92,12 @@ def delete_checkpoint(self, checkpoint_id, path):
         with self.perm_to_403():
```
etejedor, line 92: What's the rational behind making the creation of the checkpoint async but leaving the deletion sync?

etejedor: And a similar comment for restore_checkpoint.

SwanContents/swancontents/filemanager/eos/fileio.py
```diff
@@ -107,12 +108,11 @@ def _get_os_path(self, path):
         else:
             return super()._get_os_path(path)
 
-    def _save_notebook(self, os_path, nb, capture_validation_error=None):
+    async def _save_notebook(self, os_path, nb, capture_validation_error=None):
         """
         Save a notebook on EOS via FUSE with the default routine.
         Plus, store a copy of the same notebook on the host machine via docker volume.
         """
-        import time
```
etejedor, line 115: Won't this removal cause a lookup error in line 138?

SwanContents/swancontents/filemanager/eos/fileio.py
```diff
@@ -132,45 +132,46 @@ def read_notebook_from_local(path, encoding='utf-8', as_version=4):
...
-                if (nb == read_nb):
+            # Use asyncio.to_thread to offload blocking I/O operations
+            await asyncio.to_thread(write_notebook_to_local, local_path, nb)
+            for i in range(local_retry):
+                read_nb = await asyncio.to_thread(read_notebook_from_local, local_path)
+                if nb == read_nb:
                     break
                 else:
-                    #time.sleep(0.1*2**i)    # Backoff on retry (100ms to 51.2s)
-                    time.sleep(0.5)
```
etejedor, line 150: Can we safely remove this sleep? Why was it there @diocas ?

SwanContents/swancontents/filemanager/eos/fileio.py
```diff
@@ -132,45 +132,46 @@ def read_notebook_from_local(path, encoding='utf-8', as_version=4):
...
+
+            try:
+                yield file
+            except:
+                raise
+
         else:
-            # Return to the default behaviour
-            super().atomic_writing(os_path, *args, **kwargs)
+            yield await asyncio.to_thread(super().atomic_writing, os_path, *args, **kwargs)
```
etejedor, line 177: Why did you introduce `yield` here? In the previous version we were not yielding.

SwanContents/swancontents/filemanager/eos/fileio.py
```diff
@@ -132,45 +132,46 @@ def read_notebook_from_local(path, encoding='utf-8', as_version=4):
...
         From the original documentation:
         wrapper around atomic_writing that turns permission errors to 403.
         Depending on flag 'use_atomic_writing', the wrapper perform an actual atomic writing or
         simply writes the file (whatever an old exists or not)"""
 
         if self.use_atomic_writing:
-            with self.perm_to_403(os_path):
-                with atomic_writing(os_path, *args, log=self.log, **kwargs) as f:
-                    yield f
+            self.perm_to_403(os_path)
```
etejedor, line 166: You removed a context manager access, but you did not compensate later with what is executed in its exit function? Same for the enter function, was something done in enter that we are not doing anymore?

Can't you keep the first-level context manager (with self.perm ...)?

SwanContents/swancontents/filemanager/eos/fileio.py
```diff
@@ -132,45 +132,46 @@ def read_notebook_from_local(path, encoding='utf-8', as_version=4):
...
+
+            # Run the synchronous atomic_writing function in a new thread
+            file = await asyncio.to_thread(None, atomic_writing, os_path, *args, **kwargs)
+
+            try:
+                yield file
+            except:
+                raise
```
etejedor, line 174: Why this construct? Why not just yielding `f`?

SwanContents/swancontents/filemanager/projects_mixin.py
```diff
@@ -62,10 +62,10 @@ def _contains_swan_folder_name(self, path):
 
         return False
 
-    def _dir_model(self, path, content=True):
+    async def _dir_model(self, path, content=True):
         """ When returning the info of a folder, add the info of the project to which it belong to (if inside a Project) """
 
-        model = super()._dir_model(path, content)
+        model = await super()._dir_model(path, content)
```
etejedor, line 68: General comment for this and similar changes in this file: here we are calling the `_dir_model` method of `HasTraits`, which is the superclass of `ProjectsMixin`. Is `HasTraits._dir_model` a coroutine? And if so, how is it possible that we were calling it before as a normal method (not awaited)?

SwanContents/swancontents/filemanager/swan_eos_filemanager.py
```diff
@@ -9,7 +9,7 @@
 import shutil
 
 
-class SwanEosFileManager(ProjectsMixin, SwanFileManagerMixin, LargeFileManager):
+class SwanEosFileManager(ProjectsMixin, SwanFileManagerMixin, AsyncLargeFileManager):
```
etejedor, line 12: Does this change in the superclass have any implication? There's no change in the private methods we define in this class, but perhaps this class is used from somewhere else and we are invoking the coroutine equivalent of methods that we used to invoke on it before?

SwanContents/swancontents/filemanager/projects_mixin.py
etejedor: > It is defined here: https://github.com/jupyter-server/jupyter_server/blob/c67a46befdcc972436150e9876b5d25d3c33ea56/jupyter_server/services/contents/filemanager.py#L750

What inheritance relationship has the class you are pointing to with your link (AsyncFileContentsManager) with HasTraits, which is the superclass you are invoking in your code here?

SwanContents/swancontents/filemanager/eos/fileio.py
diocas: I'm not sure... Actually, now that we have k8s and no longer mount a local path, we could even remove all of this... The whole method...

 (this was the code we used in the old day to save files locally in puppet nodes, in case EOS was down... This was a measure we introduced to be able to recover data if users complained...)

SwanContents/swancontents/filemanager/eos/fileio.py
diocas: Indeed, I would suggest all of this is done by comparing with upstream. Otherwise, with this change, you are removing the conversion of errors to the proper HTTP code.

And I believe in upstream they did not make this method asynchronous, as the write call is the one being yielded.

SwanContents/swancontents/checkpoints/eoscheckpoints.py
PMax5: I changed the other methods to be async as well

SwanContents/swancontents/checkpoints/eoscheckpoints.py
```diff
@@ -91,12 +92,12 @@ def delete_checkpoint(self, checkpoint_id, path):
...
             os.unlink(cp_path)
 
-    def list_checkpoints(self, path):
+    async def list_checkpoints(self, path):
         """ On notebook opening, returns a list of all available versions. """
         base = self._get_checkpoint_base(path)
 
         try:
-            files = os.listdir(base['base_path'])
+            files = await asyncio.to_thread(os.listdir, base['base_path'])
```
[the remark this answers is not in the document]
PMax5: We can use that one too, but anyio uses asyncio so it would be an additional layer. @etejedor thoughts?

SwanContents/swancontents/filemanager/eos/fileio.py
PMax5: Ah I missed it, but the method has been fully removed. The only part that was kept was the one that calls the atomic_writing context manager.

SwanContents/swancontents/filemanager/eos/fileio.py
PMax5: Remove the entire method, but the part that calls the atomic_writing context manager.

SwanContents/swancontents/filemanager/swan_eos_filemanager.py
PMax5: Yes we are invoking the coroutine equivalent of methods that we used to invoke (https://github.com/jupyter-server/jupyter_server/blob/c67a46befdcc972436150e9876b5d25d3c33ea56/jupyter_server/services/contents/largefilemanager.py#L85)

SwanContents/swancontents/filemanager/eos/fileio.py
PMax5: I thought it something else, fixed now.

SwanContents/swancontents/filemanager/eos/fileio.py
```diff
@@ -132,45 +132,46 @@ def read_notebook_from_local(path, encoding='utf-8', as_version=4):
...
         simply writes the file (whatever an old exists or not)"""
 
         if self.use_atomic_writing:
-            with self.perm_to_403(os_path):
-                with atomic_writing(os_path, *args, log=self.log, **kwargs) as f:
-                    yield f
+            self.perm_to_403(os_path)
+
+            # Run the synchronous atomic_writing function in a new thread
+            file = await asyncio.to_thread(None, atomic_writing, os_path, *args, **kwargs)
```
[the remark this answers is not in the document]
PMax5: I added the old method again, so everything is as before (to be aligned with upstream).

SwanContents/swancontents/filemanager/eos/fileio.py
PMax5: I added the old method again, so everything is as before (to be aligned with upstream).

SwanContents/swancontents/filemanager/projects_mixin.py
PMax5: The object that calls the method is of type SwanEosFileManager, so the super()._dir_model() will resolve to one of the superclasses, in this case AsyncFileContentsManager.
image

SwanContents/swancontents/checkpoints/eoscheckpoints.py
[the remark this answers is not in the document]
etejedor: @diocas you think it's worth the installation of that additional library? @PMax5 does it simplify the coding?

SwanContents/swancontents/filemanager/eos/fileio.py
```diff
@@ -107,58 +108,6 @@ def _get_os_path(self, path):
         else:
             return super()._get_os_path(path)
 
-    def _save_notebook(self, os_path, nb, capture_validation_error=None):
```
etejedor, line 110: So I understand it is fine to remove all this @diocas ?

etejedor: Confirmed by @PMax5 as a result of a conversation with @diocas  -- this was needed in the Puppet times.

SwanContents/swancontents/checkpoints/eoscheckpoints.py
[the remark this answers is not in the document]
PMax5: I believe it is the same thing in terms of code. However, it will require an additional layer, which is not needed in my opinion.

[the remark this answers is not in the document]
PMax5: I changed it now

diocas (approved):
My only comment if that the removal of the `write_notebook_to_local` method should be in a separate commit, with the explanation that this is no longer needed.
Otherwise, perfect as is!

diocas (approved):

PR #428, opened 2025-02-27
swancustomenvironments: Read output from log file in all children tabs

rodrigo-sobral:
All tabs following the first one will read output from the log file instead of reading it from the process stdout, in order to avoid having asymmetrical (or complementary) results among tabs.


SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -36,17 +38,31 @@ async def get(self):
             # The first get request will launch the execution of the makenv.sh script,
             # to create the environment
             makenv_process = self._launch_makenv()
-            SwanCustomEnvironmentsApiHandler.makenv_process = makenv_process
             log_file = open(self.LOG_FILE, "w")
+            SwanCustomEnvironmentsApiHandler.makenv_process = makenv_process
+            SwanCustomEnvironmentsApiHandler.writing_log = True
         else:
+            # Parent process has already been launched, so we will get its generated output
```
etejedor, line 45: It's not the parent process that has been launched, it is actually a child. You can say `The makenv.sh script process has already been launched, ...`

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -36,17 +38,31 @@ async def get(self):
             # The first get request will launch the execution of the makenv.sh script,
             # to create the environment
             makenv_process = self._launch_makenv()
-            SwanCustomEnvironmentsApiHandler.makenv_process = makenv_process
             log_file = open(self.LOG_FILE, "w")
+            SwanCustomEnvironmentsApiHandler.makenv_process = makenv_process
+            SwanCustomEnvironmentsApiHandler.writing_log = True
```
etejedor, line 43: I'm thinking that it would be clearer if we place this code (and what is after the if else) in a helper method called `_process_first_get`, and what is in the else part in another helper method called `_process_subsequent_get`.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -36,17 +38,31 @@ async def get(self):
...
-                with open(self.LOG_FILE, "r") as log_file:
+            if ret_code is None or ret_code != 0:
+                log_file = open(self.LOG_FILE, "r")
+                if ret_code is None:
+                    # The process is still running, so we will stream all its output
+                    while SwanCustomEnvironmentsApiHandler.writing_log:
+                        # If the log file is being written, we will read it and send its content
+                        line = log_file.readline()
+                        if not line:
+                            # Wait for the next line to be written
```
etejedor, line 55: This is actually not waiting at all, right? It is just yielding, otherwise we would burn CPU in a busy wait.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -36,17 +38,31 @@ async def get(self):
...
+            SwanCustomEnvironmentsApiHandler.writing_log = True
         else:
+            # Parent process has already been launched, so we will get its generated output
             ret_code = makenv_process.poll()
-            if ret_code is not None and ret_code != 0:
-                # If the process has finished with error, show the content of the log file
-                with open(self.LOG_FILE, "r") as log_file:
+            if ret_code is None or ret_code != 0:
+                log_file = open(self.LOG_FILE, "r")
+                if ret_code is None:
```
etejedor, line 49: This is a bit convoluted, you are checking again a condition you already checked. Do we need a different behaviour depending on whether the makenv process is still running or not, and whether it has finished with error or not? Don't we just need to output its log?

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -31,37 +33,28 @@ async def get(self):
...
+            await self._process_subsequent_get(makenv_process)
+            self.finish()
+            return
 
         # The first and any subsequent get requests will read from the process
         # and stream out its output
-        for line in iter(makenv_process.stdout.readline, b""):
-            line = line.decode("utf-8")
-            if log_file:
+        with open(self.LOG_FILE, "w") as log_file:
```
etejedor, line 45: Everything after the if else can also go in `_process_first_get`, right?

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -31,37 +33,28 @@ async def get(self):
         self.set_header("Content-Type", "text/event-stream")
         makenv_process = SwanCustomEnvironmentsApiHandler.makenv_process
         
-        log_file = None
         if makenv_process is None:
-            # The first get request will launch the execution of the makenv.sh script,
-            # to create the environment
-            makenv_process = self._launch_makenv()
```
etejedor, line 38: You could maybe keep this call as it is, just to get the makenv process. And after that you launch the reading (_process_first_get).

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -75,7 +68,33 @@ def _launch_makenv(self) -> subprocess.Popen:
...
+        SwanCustomEnvironmentsApiHandler.writing_log = True
+        return makenv_process
+
+    async def _process_subsequent_get(self, makenv_process) -> None:
+        """
+        The makenv.sh script process has already been launched, 
+        so we will just read its output from the log file
+        """
+        ret_code = makenv_process.poll()
+        if ret_code is None or ret_code != 0:
```
etejedor, line 84: So if the process has finished with no error (ret_code == 0), we do not even show any logs, we just go to jupyterlab, right?

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -31,37 +33,28 @@ async def get(self):
...
 
-        if log_file:
-            log_file.close()
+        SwanCustomEnvironmentsApiHandler.writing_log = False
 
         self.finish()
+        return
 
     def _launch_makenv(self) -> subprocess.Popen:
+        """Launches the makenv.sh script as an async subprocess."""
```
etejedor, line 57:
```suggestion
        """Launches the makenv.sh script as a subprocess."""
```

SwanCustomEnvironments/swancustomenvironments/serverextension.py
rodrigo-sobral: Exactly, the users don't need to get the logs because the page acts just like a middle layer, redirecting them automatically to JupyterLab.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -31,37 +33,17 @@ async def get(self):
         self.set_header("Content-Type", "text/event-stream")
         makenv_process = SwanCustomEnvironmentsApiHandler.makenv_process
         
-        log_file = None
         if makenv_process is None:
-            # The first get request will launch the execution of the makenv.sh script,
```
etejedor, line 36: Would be nice to keep this comment?

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -31,37 +33,17 @@ async def get(self):
...
-            # to create the environment
             makenv_process = self._launch_makenv()
-            SwanCustomEnvironmentsApiHandler.makenv_process = makenv_process
```
etejedor, line 39: You can just do this assignment above, instead of doing it in the helper function? The rational is that _process_first_get is just in charge of reading output and sending to frontend, just like process_subsequent_get.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -75,7 +57,44 @@ def _launch_makenv(self) -> subprocess.Popen:
...
+        """
+        ret_code = makenv_process.poll()
+        if ret_code is None or ret_code != 0:
+            with open(self.LOG_FILE, "r") as log_file:
+                # The process is still running, so we will stream all its output
+                while SwanCustomEnvironmentsApiHandler.writing_log:
+                    # If the log file is being written, we will read it and send its content
+                    line = log_file.readline()
+                    if not line:
+                        # Wait for the next line to be written
```
etejedor, line 91: Again, here we do not really wait (it's sleep 0). We release control so that other coroutines can be executed.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -31,37 +33,17 @@ async def get(self):
...
-            if log_file:
-                log_file.write(line)
-            await self._send_line(line)
-
-        if log_file:
-            log_file.close()
+            await self._process_subsequent_get(makenv_process)
 
         self.finish()
+        return
```
etejedor, line 43: This return is not necessary,

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -75,7 +59,41 @@ def _launch_makenv(self) -> subprocess.Popen:
...
+                line = line.decode("utf-8")
+                log_file.write(line)
+                await self._send_line(line)
+
+        SwanCustomEnvironmentsApiHandler.writing_log = False
+
+    async def _process_subsequent_get(self, makenv_process) -> None:
+        """
+        Reads from a log file and sends to the frontend the output generated by the makenv.sh process.
+	    This is done by any subsequent get request following the initial one, which triggers the makenv.sh process.
```
etejedor, line 82: Indentation.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -75,7 +59,46 @@ def _launch_makenv(self) -> subprocess.Popen:
...
+            with open(self.LOG_FILE, "r") as log_file:
+                while SwanCustomEnvironmentsApiHandler.writing_log:
+                    # If the log file is being written, we will read it and send its content
+                    line = log_file.readline()
+                    if line:
+                        await self._send_line(line)
+        elif ret_code != 0:
+            # The process has failed, so we will read the log file and send all its content
+            with open(self.LOG_FILE, "r") as log_file:
+                lines = "".join(log_file.readlines())
```
etejedor, line 97: Not super likely, but better to read and send line by line, in case the log file is very big? Just to be on the safe side.

etejedor (approved):

PR #431, opened 2025-03-06
swancustomenvironments: Output of creation process refactor

rodrigo-sobral:
Since SparkConnector was updated with the inclusion of the changes for SwanCustomEnvironments, the "hack section" is no longer needed.

In spite of sharing the same result, sequent tabs don't show up the same lines at the same time, so the mechanism to make the creation process output synchronized was rethought.
A asynchronous subprocess is initiated instead, and its output is directly streamed to a log file. From that moment on, all tabs (including the first one making the request) will read from the same file until the process is finished or fails. No matter if the a line is actually read or not, the control is always released, so the processes can run in parallel without getting stuck.


SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -57,50 +50,24 @@ def _launch_makenv(self) -> subprocess.Popen:
...
-        SwanCustomEnvironmentsApiHandler.writing_log = True
-
-        # The first and any subsequent get requests will read from the process
-        # and stream out its output
         with open(self.LOG_FILE, "w") as log_file:
-            for line in iter(makenv_process.stdout.readline, b""):
-                line = line.decode("utf-8")
-                log_file.write(line)
+            # Create the environment asynchronously and send its output to the log file
+            return await asyncio.create_subprocess_exec(self.makenv_path, *arguments, stdout=log_file, stderr=log_file)
```
etejedor, line 55: Why does this need to become an `asyncio.subprocess.Process`? Doesn't it work the same with a regular `subprocess.Popen`? The latter is also asynchronous in the sense that it returns the process handle immediately and continues execution.

SwanCustomEnvironments/swancustomenvironments/serverextension.py
```diff
@@ -57,50 +50,24 @@ def _launch_makenv(self) -> subprocess.Popen:
...
+
+    async def _process_log_stream(self, makenv_process: asyncio.subprocess.Process) -> None:
+        """Reads from the log file and streams output to the client."""
+        with open(self.LOG_FILE, "r") as log_file:
+            while True:
+                line = log_file.readline()
+                if not line and makenv_process.returncode is not None:
+                    break  # Process finished, no more logs to read
                 await self._send_line(line)
```
etejedor, line 64: Here you are sending an empty line in case no line can be read from the file but the returncode is still None (the makenv process is running).

etejedor: I see, you protect for that case in `_send_line`, nevermind.

PR #434, opened 2025-04-03
Fixes from Diogo suggestions

rodrigo-sobral:
Automatically scroll to the `output-box` height

### DIFF
--- a/SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
+++ b/SwanCustomEnvironments/swancustomenvironments/templates/customenvs.html
@@ -73,6 +73,9 @@ data-base-url="{{base_url | urlencode}}"
     const urlParams = new URLSearchParams(window.location.search);
     const file = `/${urlParams.get('file') || ""}`;
 
+    // Scroll to the output-box height
+    window.scrollTo(0, outputBox.scrollHeight);
+
     const get_jh_auth_header = () => {
         const xsrf = document.cookie.match(/\b_xsrf=([^;]*)\b/);
         return xsrf ? { 'X-XSRFToken': xsrf[1] } : {};


SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -226,6 +226,15 @@ if [ ! -d "${GIT_REPO_PATH}" ]; then
     mv ${TMP_REPO_PATH} ${GIT_HOME}
 fi
 
+# Define the global gitignore file location
```
etejedor, line 229: General comment: this git ignore feature should be implemented elsewhere, possibly in some layer of the user image, and it should not be tied to custom environments. The same issue can happen if we are using JupyterLab with LCG and the git extension -- i.e. the git extension is not tied exclusively to the custom environments.

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -226,6 +226,15 @@ if [ ! -d "${GIT_REPO_PATH}" ]; then
...
+GLOBAL_GITIGNORE="/home/$USER/.gitignore_global"
```
etejedor, line 230: Does this setting allow the user to have .gitignore files in their git repos that override whatever is set by the global gitignore?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -226,6 +226,15 @@ if [ ! -d "${GIT_REPO_PATH}" ]; then
...
+
+# Check if Git global excludesfile is set
+if [[ "$(git config --global --get core.excludesfile)" != "$GLOBAL_GITIGNORE" ]]; then
```
etejedor, line 233: So this can collide if the user already set a global gitignore file?

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
rodrigo-sobral: Yes, local `.gitignore` (per repo) takes priority to what's globally ignored. e.g. if `.gitignore_global` has `*.sys.*`, you can create a `.gitignore` in your local repo and use `!important.sys.txt` for overwriting your global config

SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
rodrigo-sobral: Yes, the global gitignore config can only be pointing to one file at a time

etejedor: But then as per this logic we could be adding content to a global gitignore file of the user that is located in their EOS?

Is it possible to just set this variable to some value, and then the user can have e.g. a per-user gitignore setting in their HOME that they do? And if they do nothing, our variable takes precedence. If they do something, our variable is ignored.

diocas (approved):

etejedor (approved):

PR #438, opened 2025-05-22
swancustomenvironments: Add a 10s timeout on cloning

rodrigo-sobral:
As ATS clusters do not access to the internet, causing repository clones from Github to fail, a timeout needs to be explicitly done to lead to an error that outputs a informative message to the user


SwanCustomEnvironments/swancustomenvironments/scripts/makenv.sh
```diff
@@ -140,7 +140,7 @@ if [[ "$REPOSITORY" =~ $REPO_GIT_PATTERN ]]; then
     # Otherwise, use the existing one in the SWAN_projects folder
     if [ ! -d "${GIT_REPO_PATH}" ]; then
         _log "Cloning the repository from ${REPOSITORY}..."
-        git clone $REPOSITORY -q "${TMP_REPO_PATH}" 2>&1
+        timeout 10s git clone $REPOSITORY -q "${TMP_REPO_PATH}" 2>&1
```
etejedor, line 143: 10 seconds is perhaps too short? What is the default timeout that the git command itself has?
Also, does this change by itself improve the error message that the user sees?

rodrigo-sobral: I don't think the git command has any type of internal timout on itself. But the https request does. From my personal experience, I would say its current timeout is about 20s, and when it reaches that point, it results on a error page with that orange button to go back to the form (the request timed out), but no message is printed in the box about what happened. Timing out the command itself, makes it to generate an error that leads to a message being sent to the frontend and a process stopping. Where the different result as before and after the change

image

etejedor: It just seems to me 10s is too short. There can be git repos with big content that take more than 10s to clone.

When do you not use the timeout prefix, what is the return code of the git command when it timeouts by itself? Is it zero?

etejedor (approved):

PR #441, opened 2025-07-03
swandask: Fix error when connecting to HTCondor clusters on `bleeding edge`

rodrigo-sobral:
In spite of fixing the bugs presented in the `bleeding edge` software stack when connecting to HTCondor cluster, the 403 error still persists.

```
2025-07-03 15:08:42,651 - SwanDask - INFO - Running SwanDask on port 41373 with base url /user/rhenriqu/
[W 2025-07-03 15:08:43.470 ServerApp] 403 GET /user/rhenriqu/dask/clusters?1751548120525 (rhenriqu@127.0.0.1) 2725.25ms
[W 2025-07-03 15:08:43.471 ServerApp] 403 GET /user/rhenriqu/dask/clusters?1751548120543 (rhenriqu@127.0.0.1) 2694.93ms
```

### DIFF
--- a/SwanDask/swandask/app.py
+++ b/SwanDask/swandask/app.py
@@ -12,6 +12,8 @@ from tornado import ioloop, web
 # Otherwise, we would need to configure the templates
 from jupyter_server.base.handlers import JupyterHandler, APIHandler
 JupyterHandler.write_error = APIHandler.write_error
+from traitlets.config import Configurable
+from jupyter_server.auth import IdentityProvider
 
 class WebApp:
     pass
@@ -46,11 +48,15 @@ def main():
     # Prevent 403 errors when querying the dashboard server
     _set_dashboard_whitelist()
 
+    # Create a dummy Configurable parent for the IdentityProvider
+    identity_provider = IdentityProvider(parent=Configurable())
+
     # If no remote access allowed, Jupyter will check if we're serving from https://localhost
     app = web.Application(
         base_url=args.base_url,
         allow_remote_access=True,
         cookie_secret=urandom(32),
+        identity_provider=identity_provider
     )
 
     server_app = WebApp()


SwanDask/swandask/app.py
```diff
@@ -36,6 +37,14 @@ def main():
     parser.add_argument("--base_url", default="/", action="store", dest="base_url")
     args = parser.parse_args()
 
+    jupyter_runtime_dir = os.path.join(os.environ.get("JUPYTER_RUNTIME_DIR", "/tmp"), "jupyter_cookie_secret")
```
etejedor, line 42: Is this a secret that is generated automatically by jupyter at this path?

rodrigo-sobral: Exactly, that file is automatically generated by jupyter_server as you can confirm in here: https://github.com/jupyter-server/jupyter_server/blob/56e2478a728ff292d8270e62d27dd50c316ee6b7/jupyter_server/serverapp.py#L1146,
and as we define `JUPYTER_RUNTIME_DIR` like this: https://github.com/swan-cern/jupyter-images/blob/main/swan/scripts/before-notebook.d/02_environment.sh#L18

etejedor: We seem to be passing that secret for another tornado web app here:

https://github.com/swan-cern/jupyterhub-extensions/blob/master/SwanNotificationsService/swannotificationsservice/app.py#L26

in the notifications extensions. There we just pass a randomly generated value on the spot. Is that enough? @diocas ?

diocas: I have no idea, sorry... But please be aware that the notifications service runs independently.

But I don't understand why we are doing auth in here (?). Isn't the jupyter-proxy in front of it doing that already?

etejedor: It seems that the move to jupyter_server 2.X in the LCG releases now requires this argument here.

etejedor: The issue here is that we tied SwanDask to the LCG release environment, because we needed Dask from it. So we are subject to jupyter_server updates from the LCG stack, while the mamba env still runs v1.X.

etejedor: Wait @diocas , we are already using jupyter_server 2.X, more precisely 2.15.0, in our conda environment in the user image. So our jupyter server and all the old UI stuff is already working with 2.X. The only piece that was left behind was SwanDask, because it was configured with the environment of the LCG release, which so far had (until LCG 107) jupyter_server 1.X, and now in bleeding edge it is 2.X.

PR #449, opened 2026-01-22
swancontents: Pass the value of require_hash down the stack

PMax5:
In more recent versions of jupyter_server, require_hash argument no longer accepts the value None. After upgrading to JupyterLab 4.5.2 (which also upgraded other jupyter components), the jupyter_server forced to set the hash and hash_algorithm parameters and so `require_hash` must be true in order not to show an error regarding missing parameters to the user.

The problem was that we were not passing such argument down the function call and so it was setting it to `False` as default, thus showing the aforementioned error about the missing parameters.


rodrigo-sobral (approved):

PMax5:
This PR will stay on hold until JupyterHub 5 is merged an deployed in production

PR #479, opened 2026-04-23
Add a simple extension to integrate with ml.cern.ch (Kubeflow)

tomasr8:
Just putting it here since it's ready. I'm planning to undraft this only once we have the token exchange configured.

This adds a simple sidebar panel to the JupyterLab UI which shows the status of Kubeflow pipeline runs.

This extension uses the kfp client to communicate with Kubeflow. It reads an API token from the file system (location configurable).

mlcern


tomasr8:
This should be reviewable now. When testing make sure to use the latest swan-cern chart (it enables token exchange with kubeflow)

rodrigo-sobral (approved):

PMax5 (approved):

PR #482, opened 2026-05-06
swannotifications: Remove unused extension

PMax5:
Remove the SwanNotifications extension as it is no longer used in the user image


tomasr8 (approved):
Looks good. I'll merge it after the intervention tomorrow.

rodrigo-sobral (approved):

PR #487, opened 2026-05-13
sparkconnector: Add listener when pressing <Enter> to connect

rodrigo-sobral:


SparkConnector/src/components/authenticate.tsx
```diff
@@ -42,6 +42,11 @@ export const Authenticate = observer(() => {
           onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
             setPassword(event.target.value);
           }}
+          onKeyPress={(event: React.KeyboardEvent<HTMLInputElement>) => {
```
tomasr8: Looks like `onKeyPress`/`keypress` is deperecated: https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event
Maybe we can use `onKeyDown` or similar?

rodrigo-sobral: Fixed, thx!

tomasr8 (approved):

PR #490, opened 2026-06-12
swancontents: Fix preferred_dir interaction with SWAN_projects

tomasr8:
SWAN_projects is a special folder in both the classic UI and JupyterLab. Users cannot create and/or rename folders with that name. This was previously enforced with some logic relying on `root_dir` where the hidden assumption was that `root_dir` matches $HOME. This is not the case in JupyterLab where `root_dir` is set to `/eos` and `preferred_dir` is used in conjunction.

The effect was that in JupyterLab it is not possible to rename files inside SWAN_projects. It is also not possible to download files from anywhere.

The fix is to define a new propety `swan_home` which always resolves to the user home.

### DIFF
--- a/SwanContents/swancontents/filemanager/swan_eos_filemanager.py
+++ b/SwanContents/swancontents/filemanager/swan_eos_filemanager.py
@@ -46,7 +46,7 @@ class SwanEosFileManager(ProjectsMixin, SwanFileManagerMixin, AsyncLargeFileMana
         Define the root path for tornado StaticFileHandler object
         This is necessary to open files from other users (for sharing tab)
         """
-        if self.root_dir.startswith("/eos/"):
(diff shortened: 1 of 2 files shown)


SwanContents/swancontents/filemanager/swan_eos_filemanager.py
```diff
@@ -46,7 +46,7 @@ def _files_handler_params_default(self):
         Define the root path for tornado StaticFileHandler object
         This is necessary to open files from other users (for sharing tab)
         """
-        if self.root_dir.startswith("/eos/"):
+        if self.swan_home.startswith("/eos/"):
```
tomasr8, line 49: This fixes file downloads in JupyterLab

rodrigo-sobral (approved):

PR #491, opened 2026-06-30
swanintro: Remove unused extension

rodrigo-sobral:
After checking all our repositories, we didn't find any references to SwanIntro, leading us to the conclusion it became a deprecated extension over time.


tomasr8:
From what I can see, it's actually still used. When a user does not have `SWAN_projects/.intro` file (or the version number stored inside is different) it shows a welcome modal.

rodrigo-sobral:
I noticed that, but when I tried to remove that `.intro` file and check if the modal was actually being shown, it didn't.

tomasr8 (approved):

PR #494, opened 2026-09-01
ci: add ci runner for unit tests and ruff check

icefox97:


.github/workflows/ci.yml
```diff
@@ -62,3 +62,19 @@ jobs:
         env:
           PACKAGE: ${{ matrix.package }}
         run: uv build ${PACKAGE}
+
+      - name: Lint ${{ matrix.package }} 🧹
+        env:
+          PACKAGE: ${{ matrix.package }}
+        run: cd ${PACKAGE} && uvx ruff@0.15.20 check .
```
tomasr8: Can we use something that dependabot will be able to automatically update? Maybe the [ruff-action](https://github.com/astral-sh/ruff-action)?

.github/workflows/ci.yml
```diff
@@ -62,3 +62,19 @@ jobs:
...
+      - name: Test ${{ matrix.package }} 🧪
+        env:
+          PACKAGE: ${{ matrix.package }}
+        run: |
+          cd ${PACKAGE}
+          if grep -q '\[tool\.pytest\.ini_options\]' pyproject.toml; then
+            uv run --group dev pytest -v
+          else
+            echo "No tests configured for ${PACKAGE}, skipping."
+          fi
```
tomasr8: pytest returns exit code 5 when it doesn't find any tests, that might be better to test for than the pyproject config (which could be missing)

SwanCustomEnvironments/pyproject.toml
```diff
@@ -75,5 +75,12 @@ before-build-npm = [
 ]
 before-build-python = ["jlpm clean:all"]
 
+[dependency-groups]
+dev = ["pytest>=8", "pytest-asyncio>=0.23", "jupyter_server>=2.0.1,<3"]
```
tomasr8: I don't think we need `jupyter_server` here since it's already a regular dependency

SwanCustomEnvironments/pyproject.toml
```diff
@@ -75,5 +75,12 @@ before-build-npm = [
...
+
+[tool.pytest.ini_options]
+asyncio_mode = "auto"
+testpaths = ["swancustomenvironments/tests"]
```
tomasr8: I believe this is the default path, so I think it can be omitted.

.github/workflows/ci.yml
```diff
@@ -62,3 +62,16 @@ jobs:
         env:
           PACKAGE: ${{ matrix.package }}
         run: uv build ${PACKAGE}
+
+      - name: Lint ${{ matrix.package }} 🧹
+        uses: astral-sh/ruff-action@d0a0e814ec17e92d33be7d24dd922b479f1bcd38 # v1
```
tomasr8: The latest is v4.1.0: https://github.com/astral-sh/ruff-action/releases

.github/workflows/ci.yml
```diff
@@ -62,3 +62,16 @@ jobs:
...
+        with:
+          version: "0.15.20"
```
tomasr8: Can we use the latest ruff version as well here?

SwanCustomEnvironments/swancustomenvironments/tests/test_ci_sanity.py
```diff
@@ -0,0 +1,2 @@
+def test_ci_runs():
+    assert True
```
tomasr8: Let's keep this out of the PR

SwanCustomEnvironments/pyproject.toml
```diff
@@ -78,6 +78,12 @@ before-build-python = ["jlpm clean:all"]
 [dependency-groups]
 dev = ["pytest>=8", "pytest-asyncio>=0.23"]
 
+[tool.ruff.lint]
```
tomasr8: It'd be nice to have one global `ruff.toml` instead having one config for reach package (they will be most likely identical anyway)

ruff.toml
```diff
@@ -0,0 +1,5 @@
+[lint]
+ignore = [
+    "I001",    # TODO: fix import ordering across all packages
```
tomasr8:
```suggestion
    "I001",     # TODO: fix import ordering across all packages
```

tomasr8 (approved):

PR #497, opened 2026-09-11
swancontents: Fix gallery and EOS downloads in JupyterLab

cameron-lemon:
Currently two download methods are failing : 1) in JupyterLab sessions, opening a
notebook from the SWAN gallery in the launcher using the "Open in SWAN"
button fails ("Failed to download notebook" panel appears), and 2) opening
a notebook from CERNBox in your home folder (right click, "Open in SWAN") and
selecting the new JupyterLab session fails with "Error downloading Project:
Network response was not ok".

The fix is related to the root_dir being different for the old UI and the JupyterLab
UI. Following #492, which fixed this for the git clone branch of download() by
implementing a new swan_home function, we update the places root_dir was
still being used instead of swan_home.

This is independent of #493, which changes path resolution in get(). This PR
only changes where download() writes.

### Testing
Dev cluster with a patched image, both UIs:

| Case | Classic | JupyterLab |
|---|---|---|
| SWAN gallery | ✅ | ✅ (failed before) |
| Own EOS file | ✅ | ✅ (failed before) |
| EOS file outside home (`/eos/project/...`) | ✅ | ✅ |

### Known behaviour difference (not changed here)
For "Open in SWAN" on EOS files, the two UIs open the file in different places.
Left as is for now, to be revisited with upcoming CERNBox sharing.


SwanContents/swancontents/serverextension/handlers/__init__.py
tomasr8, line 1: Why the file mode change?

SwanContents/pyproject.toml
```diff
@@ -1,3 +1,6 @@
 [build-system]
 requires = ["jupyter_packaging~=0.7.9", "setuptools>=40.8.0", "wheel"]
 build-backend = "setuptools.build_meta"
+
+[dependency-groups]
+dev = ["pytest>=8"]
```
tomasr8: newline ;)

SwanContents/swancontents/serverextension/handlers/__init__.py
cameron-lemon: This was to fix an easy ruff complaint about executable with no shebang.
Reverted to keep this PR clean and ruff clean-up can come separately.

SwanContents/pyproject.toml
cameron-lemon: Fixed for here and ruff.toml

SwanContents/swancontents/filemanager/projects_mixin.py
```diff
@@ -1,7 +1,7 @@
 
 from traitlets import HasTraits, Unicode
 from tornado import web
-import os, io, shutil, subprocess, tempfile, requests
+import os, io, shutil, subprocess, tempfile, requests, zipfile
```
tomasr8, line 4: Good we have ruff now to catch these 😆

tomasr8 (approved):

SwanContents/swancontents/serverextension/handlers/__init__.py
tomasr8: Fine by me if you want to fix it, I was mainly curious

PR #498, opened 2026-09-17
swancontents: Fix ruff complaints for unused imports and variables

cameron-lemon:
Fix F401 import errors either unused or re-exported in __init__.py
files and ruff wants them to be declared. has_voila check now
removes the unused voila import via importlib vcheck

Fix F841 removing some unused variables and also an unused URL
check in proj_url_checker for forbidden sequences that was
never correctly implemented (testing the list instead of the match).
Could consider implementing it but most of the sequences were single
chars already not allowed, and the space and @ sequences might
break existing filepaths and gitlab URLs with credentials.

Remove ruff ignores


SwanContents/swancontents/swanclassic/handlers/notebookviewer.py
```diff
@@ -8,14 +8,9 @@
...
-has_voila = False
-try:
-    import voila
-
-    has_voila = True
-except:
-    pass
+import importlib.util
 
+has_voila = importlib.util.find_spec("voila") is not None
```
tomasr8: This changes the semantics of the import check. It's probably fine, but I'd really like to avoid such changes in this part of the codebase for the sake of making ruff happy. Perhaps you could just #noqa the line instead?

cameron-lemon: Sure, reverted the change and added the comment

tomasr8 (approved):

PR #499, opened 2026-09-17
swancontents: Fix safe Ruff syntax issues

cameron-lemon:
Ruff safe fixes for modernising the syntax, mostly updates from
having to be compatible with Python 2, no behavioural changes.
Regex updated to raw strings to avoid future SyntaxError.

Corresponding ruff.toml ignores removed

Overlaps with #498 as both remove entries from ruff.toml. I will
rebase this branch after #498 merge.


tomasr8 (dismissed):

tomasr8:
Could you fix the conflict?

